PR #9, opened 2021-06-04
Fix for the ceSimReco crash in recent CI jobs.

kutschke:
Recent Offline CI jobs, for example https://github.com/Mu2e/Offline/pull/474 , have failures in ceSimReco and in cEmix.

This PR fixes the failure in ceSimReco - changes in JobConfig were not propagated to Validation.

I am 99% sure that the crash in ceMix is because the default input filename for the mix-in files is now @nil.  So we need to provide real values in Valication/ceMix.fcl.

I am done for the day and can look at it tomorrow if no one else get's there first.


brownd1978:
The same fix was already merged.

PR #23, opened 2021-06-15
POMS campaign

soleti:
This PR contains:

- The POMS configuration files for the CORSIKA part of the MDC2020 campaign
- The Perl libraries needed by generate_fcl to run on the grid node
- An updated version of the CORSIKA input file with magnetic field updated to 06/01/2024.


brownd1978 (approved):
most of these files were generated, ie not primary sources.  Can they be regenerated from the primary sources?  If so I suggest in future to save just the primary source and any script needed to generate them as opposed to the output itself.

soleti:
Are you referring to the perl libraries? Those are needed because the perl package autodie used by generate_fcl is not available on the worker nodes. The other files were not generated.

brownd1978:
got it.  Thanks for the explanation.
I can see this being a potential maintenance issue.  Maybe we can ask scd to make that module available on the workers.  Lets discuss tomorrow in the meeting.

kutschke:
There are two ways to proceed.  One is to put the perl modules in the container that hosts all of our grid jobs.  The other is UPS.  Both work for me.  My instinct is to pose the problem to them and let them make the call.

Rob

From: David Brown ***@***.***>
Reply-To: Mu2e/Production ***@***.***>
Date: Tuesday, June 15, 2021 at 6:39 PM
To: Mu2e/Production ***@***.***>
Cc: Subscribed ***@***.***>
Subject: Re: [Mu2e/Production] POMS campaign (#23)


got it. Thanks for the explanation.
I can see this being a potential maintenance issue. Maybe we can ask scd to make that module available on the workers. Lets discuss tomorrow in the meeting.

PR #34, opened 2021-06-18
CI test - please ignore

ryuwd:
Manual test to see how FNALbuild deals with repos other than Mu2e/Offline


ryuwd:
FNALbuild is missing some required permissions that need to be granted first

PR #53, opened 2021-07-03
Add splitting steps

brownd1978:


soleti:
Could you also update the JPG and the PDF?

PR #57, opened 2021-07-07
Campaigns fixes

soleti:
This PR includes new versions of the beam and CORSIKA campaigns.

### DIFF
--- a/CampaignConfig/mdc2020_beam.ini
+++ b/CampaignConfig/mdc2020_beam.ini
@@ -2,7 +2,7 @@
 experiment = mu2e
 poms_role = production
 name = MDC2020f_beam
-campaign_stage_list = pot_fcl, pot, beam_cat_fcl, beam_cat, neutrals_cat_fcl, neutrals_cat, beam_resampler_fcl, beam_resampler, neutrals_resampler_fcl, neutrals_resampler, beamflash_cat_fcl, beamflash_cat, earlybeamflash_cat_fcl, earlybeamflash_cat, ipastops_cat_fcl, ipastops_cat, neutralsflash_cat_fcl, neutralsflash_cat, earlyneutralsflash_cat_fcl, earlyneutralsflash_cat, muonstops_fcl, muonstops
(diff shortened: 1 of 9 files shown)


CampaignConfig/mdc2020_beam.ini
```diff
@@ -2,15 +2,15 @@
...
+campaign_stage_list = pot_fcl, pot, beam_cat_fcl, beam_cat, neutrals_cat_fcl, neutrals_cat, mubeam_resampler_fcl, mubeam_resampler, elebeam_resampler_fcl, elebeam_resampler, neutrals_resampler_fcl, neutrals_resampler, elebeamflash_cat_fcl, elebeamflash_cat, earlyelebeamflash_cat_fcl, earlyelebeamflash_cat, mubeamflash_cat_fcl, mubeamflash_cat, earlymubeamflash_cat_fcl, earlymubeamflash_cat, ipastops_cat_fcl, ipastops_cat, targetstops_cat_fcl, targetstops_cat, neutralsflash_cat_fcl, neutralsflash_cat, earlyneutralsflash_cat_fcl, earlyneutralsflash_cat, muonstops_fcl, muonstops
 
 [campaign_defaults]
 vo_role=Production
 software_version=Offline
 dataset_or_split_data=None
 cs_split_type=None
 completion_type=complete
-completion_pct=95
+completion_pct=100
```
brownd1978, line 13: Does this mean that even 1 failed job will prevent the next stage from running?

brownd1978 (approved):

CampaignConfig/mdc2020_beam.ini
soleti: This is something I want to check because I am not sure. With 95% it was starting the next stage event if the remaining 5% was still running. With 100% it should start only when all the jobs ended (successfully or not). But if we have e.g. 50% of failed jobs we might not want to go to the following stage automatically... I will ask Marc what's the best way to configure this.

PR #64, opened 2021-07-15
add Offline/ to beginning of geometry file in fcl, remove a redundant…

HCasler:
… include of fcl/StandardServices.fcl


kutschke:
Tested in Offline using PR#529 and the "test with" feature.  All tests passed.

PR #65, opened 2021-07-15
more include shifts

rlcee:
fix a few more missed include shifts


kutschke:
I have made a dummy PR in Offline.  After it completes it's initial CI build. I will tweak it and the do a new CI build that tests this PR on Production.  If all goes well I will merge later this evening.

kutschke (approved):
This was validated on using Offline PR 532 .
It passed validation.

PR #67, opened 2021-07-21
Campaign refactoring

soleti:
This PR adds the MuStopPileup stage to the beam campaign and fixes few bugs.


CampaignConfig/mdc2020_beam.cfg
```diff
@@ -698,6 +723,77 @@ job_setup.getconfig = True
 job_setup.multifile = False
 job_setup.setup_local = True
 
+[stage_muminusstops_cat_fcl]
```
brownd1978, line 726: I thought we agreed not to make this explicit, since the real concatenation is done in the split and this is just copying 1 file -> 1 file.  Or maybe I'm missing something?

soleti: I thought we agreed we were going to not do the concatenation in the split anymore, but maybe I misunderstood. If that's the case i'll remove the stage.

CampaignConfig/mdc2020_corsika.cfg
```diff
@@ -188,20 +195,22 @@ executable.arg_6 = --events-per-job=65000
...
 job_output_2.dest = %(outdir_dts)s/CosmicCORSIKA/%(desc)s/art
+job_output_1.dest = %(logdir_bck)s/%(stage_name)s/%(desc)s/tbz/
 
 global.artRoot_dataset = dts.%(submitter)s.CosmicCORSIKA.%(desc)s.art
 
 job_setup.postscript = samweb file-lineage parents `basename ${fname}` > parents.txt
 
 job_setup.getconfig = True
 submit.n_files_per_job = 1
+submit.memory = 4000MB
```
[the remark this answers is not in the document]
soleti: I tried with 3000MB and it got held, so I used 4000MB. But I haven't optimized it...

CampaignConfig/mdc2020_beam.cfg
soleti: Ok, I removed that.

PR #81, opened 2021-08-09
Mixing stage and automatic MaxEventsToSkip

soleti:
This PR includes a NoPrimaryMix stage in the POMS campaign and the automated calculation of the `MaxEventsToSkip` factor for resampling stages.


CampaignConfig/mdc2020_beam.cfg
```diff
@@ -733,15 +741,22 @@ global.stage_name = MuStopPileup
...
+job_setup.prescript_5 = let nskip=nevts/nfiles
+job_setup.prescript_6 = echo '#include "Production/JobConfig/pileup/MuStopPileup.fcl"' >> template.fcl
+job_setup.prescript_7 = echo 'physics.filters.TargetStopResampler.mu2e.MaxEventsToSkip: '${nskip} >> template.fcl
+job_setup.prescript_8 = let njobs=%(njobs)s/10
 
 executable.name = true
 executable_2.name = generate_fcl
 executable_2.arg_4 = --events-per-job=400000
-executable_2.arg_5 = --njobs=50
+executable_2.arg_5 = --njobs=$njobs
```
brownd1978, line 755: This seems to be setting all the jobs (mustop resampler, neutrals resampler, ...) to the same number of jobs.  We want those in the ratio we discussed before.  Or did I miss that the njobs variable is being overwritten?

soleti: This `$njobs` corresponds to the variable defined just above `job_setup.prescript = let njobs=%(njobs)s/100` which where `%(njobs)s` is the number of POT jobs. It's probably confusing to use the same name for INI and bash variable... But I think is correct right?

brownd1978: I was just looking at the snippets of changed lines, so it's missing some context.  As long as the ratios of the numbers of jobs ends up as we discussed it's fine.  Can you run a test to verify what ratios are produced?

soleti: In order to run a test I need a tagged version because scripts like `samCountFiles.sh` must be available on cvmfs. For example I do:
```
job_setup.prescript_3 = nfiles=`/cvmfs/mu2e.opensciencegrid.org/Musings/SimJob/%(release)s/link/Production/MDC2020/POMS/samCountFiles.sh $dataset`
```

rlcee:
Muse can put this directory
/cvmfs/mu2e.opensciencegrid.org/Musings/SimJob/%(release)s/link/Production/MDC2020/POMS/samCountFiles.sh
 in your path, so you can refer to it with just
samCountFiles.sh
This would work with the local development area too.
Please see how Offline/.muse adds Offline/bin, or I'm happy to put in a PR

brownd1978 (commented):
I'm not sure what's going on with the 'empty files' messages below.  Maybe this is due to the earlier GitHub problems?

brownd1978 (approved):
It's odd but apparently GitHub lists a file permission change as an 'empty file' in the review.  Thanks Roberto for explaining that.

PR #83, opened 2021-08-11
Add artcatSplitLevel1.fcl

kutschke:
The script artcatSplitLevel1.fcl behaves like artcat.fcl but it writes the output file with splitLevel : 1.  Previous code used the default, which is splitLevel : 99.  This means fully split since our most nested data products are about 5 or 6 levels deep.

The effect of this change is that the data product payload ( the .obj branch inside the root file) unspilt unit.  

According to Chris Jone, CMS uses splitLevel : 1 if the output is intended to be ready by a production job and they make fully split files if the file is intended to be used for analysis.  The reason is that it saves memory in the running jobs at the cost of increased size on disk.

I tested a MuminusStops file from MDC2020d.  There was an increase in size on disk of 4.2%.  I have not tested the memory savings.

### DIFF
--- a/JobConfig/common/unsplit.fcl
+++ b/JobConfig/common/unsplit.fcl
@@ -1,0 +1,13 @@
+// Analog of artcat.fcl but write the output file with splitLevel : 1
+//
+// See comments in ./artcat.fcl
+//
+// Rob Kutschke, 2021
+
+#include "Offline/fcl/minimalMessageService.fcl"
+
+process_name: unsplit
+source: { module_type: RootInput }
+services: { message: @local::default_message }
+physics: { e1: [out]  end_paths: [e1] }
+outputs: { out: { module_type: RootOutput fileName: @nil splitLevel: 1 } }
--- a/JobConfig/common/unsplit.fcl
+++ b/JobConfig/common/unsplit.fcl
@@ -1,4 +1,5 @@
 // Analog of artcat.fcl but write the output file with splitLevel : 1
+// This will write each data product as an unsplit unit.
 //
 // See comments in ./artcat.fcl
 //
--- a/JobConfig/common/artcatSplitLevel1.fcl
+++ b/JobConfig/common/artcatSplitLevel1.fcl
@@ -7,7 +7,7 @@
 
 #include "Offline/fcl/minimalMessageService.fcl"
 
-process_name: unsplit
+process_name: artcatSplitLevel1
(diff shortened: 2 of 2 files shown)


kutschke:
Feel free to suggest a better name than unspilt.fcl .  I thought that artcat_splitLevel1.fcl was too unwieldy and that artcat_1 was too ambiguous.   Besides, if we adopt this widely we will just change artcat.fcl.

We can also ask  the the default be splitLevel 1 for Mu2e.

rlcee:
I wonder if you ran validation comparing the split and unsplit files?  If not, let me know and I can try it.

JobConfig/common/unsplit.fcl
```diff
@@ -0,0 +1,13 @@
...
+//
+// Rob Kutschke, 2021
+
+#include "Offline/fcl/minimalMessageService.fcl"
+
+process_name: unsplit
+source: { module_type: RootInput }
+services: { message: @local::default_message }
+physics: { e1: [out]  end_paths: [e1] }
+outputs: { out: { module_type: RootOutput fileName: @nil splitLevel: 1 } }
```
brownd1978, line 13: Since this script is hard-coded to splitlevel 1 I think that should be in the name.

kutschke: I updated it.

kutschke:
> I wonder if you ran validation comparing the split and unsplit files? If not, let me know and I can try it.

No.  I have not.  Do you mean just running val.fcl on the two files and then running valCompare?  Or something more than that. 
I will do that now.

kutschke:
@FNALbuild run build tests

kutschke:
The file used for testing was:
/cvmfs/mu2e.opensciencegrid.org/DataFiles/Validation/sim.brownd.MuminusStops.MDC2020d.001202_00000000.art

I rewrote this file was splitlevel 1.  I ran val.fcl on both files and ran valCompare on the result.  All plots are identical.

The val.fcl job ran on 73 seconds real time on the original file and in 40 seconds real time on the splitLevel 1 file.  The VmHWM reported by art was the same within a few MB for the two jobs.

The original file was:
/cvmfs/mu2e.opensciencegrid.org/DataFiles/Validation/sim.brownd.MuminusStops.MDC2020d.001202_00000000.art

brownd1978 (approved):

PR #84, opened 2021-08-18
Save only parent and direct SimParticles in pileup

brownd1978:
This change will eventually change the content of mixin samples, but won't have any immediate effect on nightly validation.


soleti (approved):
Looks good to me

PR #91, opened 2021-09-07
Number of jobs fix

soleti:
This PR fixes the number of jobs for the 1% campaign and contains a bug fix for the script that does the shuffling of the input files for the Flash output concatenation.

### DIFF
--- a/CampaignConfig/mdc2020_beam.cfg
+++ b/CampaignConfig/mdc2020_beam.cfg
@@ -308,7 +308,7 @@ executable_2.name = generate_fcl
 executable_2.arg_4 = --aux=1:physics.filters.beamResampler.fileNames:inputs.txt
 executable_2.arg_5 = --events-per-job=400000
 executable_2.arg_6 = --run-number=%(run_number)s
-executable_2.arg_7 = --njobs=%(njobs)s
+executable_2.arg_7 = --njobs=100
 executable_2.arg_8 = --embed
 executable_2.arg_9 = template.fcl
 
@@ -391,8 +391,8 @@ executable_2.name = generate_fcl
 executable_2.arg_4 = --aux=1:physics.filters.neutralsResampler.fileNames:inputs.txt
 executable_2.arg_5 = --events-per-job=400000
 executable_2.arg_6 = --run-number=%(run_number)s
-executable_2.arg_7 = --njobs=%(njobs)s
-executable_2.arg_8 = --include
+executable_2.arg_7 = --njobs=1000
+executable_2.arg_8 = --embed
 executable_2.arg_9 = template.fcl
 
 job_setup.ifdh_art = False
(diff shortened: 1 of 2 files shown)


CampaignConfig/mdc2020_beam.cfg
```diff
@@ -308,7 +308,7 @@ executable_2.name = generate_fcl
 executable_2.arg_4 = --aux=1:physics.filters.beamResampler.fileNames:inputs.txt
 executable_2.arg_5 = --events-per-job=400000
 executable_2.arg_6 = --run-number=%(run_number)s
-executable_2.arg_7 = --njobs=%(njobs)s
```
brownd1978, line 311: Can you do math in POMS?  Ie could you have a global scale variable and then have the specific number of jobs be a scaled version of that?  100 X N, where N is the percent of the production?

MDC2020/POMS/catAndShuf.sh
```diff
@@ -58,8 +58,9 @@ do
         numbers=(${sequencer//_/ })
         counter_string_old=${numbers[1]}
         counter_string=$(printf "%08d" $ifile)
-        sed -i "s/$counter_string_old/$counter_string/g" $f
-        sed -i "s/$counter_string_old/$counter_string/g" $f.json
+        sed -i "/outputs/s/$counter_string_old/$counter_string/g" $f
+        sed -i "/dh.sequencer/s/$counter_string_old/$counter_string/g" $f.json
+        sed -i "/file_name/s/$counter_string_old/$counter_string/g" $f.json
         mv "$f" `basename "${f/$counter_string_old/$counter_string}"`
```
brownd1978, line 64: Can we test this script using MDC2020i output?

CampaignConfig/mdc2020_beam.cfg
soleti: Not really. An alternative is to declare a bash variable that is then used as argument. Something like this for each stage:
`job_setup.prescript = let this_njobs=%(njobs)s*%(multiplier)s`
and then
`executable_2.arg_7 = --njobs=${this_njobs}`
which I don't like too much...

MDC2020/POMS/catAndShuf.sh
soleti: The script runs fine interactively (it was running fine also before) but I caught a bug where it was replacing also one of the input filenames (and it shouldn't). The only way to really test this is to run on the grid because only there we can verify the SAM declaration.

CampaignConfig/mdc2020_beam.cfg
brownd1978: How do you plan to scale this up for 10%, 100% production?  Doesn't it require hand-editing the configs?  That will be error-prone.
We can let this in as a bug fix, but we need to be thinking about the next set of jobs too.

soleti: I agree. My idea was to have a global variable `%(njobs)s` set to the number of POT jobs and then scale it as showed for the downstream stages.

brownd1978 (approved):

PR #93, opened 2021-09-15
Mix campaign fixes

soleti:
This PR contains a series of bug fixes in the Mix MDC2020 campaign, including setting the `DbService.purpose` to MDC2020i.

### DIFF
--- a/MDC2020/gen_Mix.sh
+++ b/MDC2020/gen_Mix.sh
@@ -80,7 +80,7 @@ echo physics.filters.MuStopPileupMixer.mu2e.MaxEventsToSkip: ${nskip_MuStopPileu
 # setup database access for SimEfficiencies
 #
 echo 'services.ProditionsService.simbookkeeper.useDb: true' >> template.fcl
-echo services.DbService.purpose: $2$4 >> template.fcl
+echo services.DbService.purpose: $2i >> template.fcl
 #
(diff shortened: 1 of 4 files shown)


MDC2020/gen_Mix.sh
```diff
@@ -80,7 +80,7 @@ echo physics.filters.MuStopPileupMixer.mu2e.MaxEventsToSkip: ${nskip_MuStopPileu
 # setup database access for SimEfficiencies
 #
 echo 'services.ProditionsService.simbookkeeper.useDb: true' >> template.fcl
-echo services.DbService.purpose: $2$4 >> template.fcl
+echo services.DbService.purpose: $2i >> template.fcl
```
brownd1978, line 83: We will trip over this hardcoded "i" eventually.   But if we make it separate, we have 3 (!) different versions: input, output, and database.  So we need a better solution before we do 10%.

soleti: Yes, I agree. Do you suggest a better solution? I can add an argument, but as you say we end up with 3 different versions... Not sure.

brownd1978: No I don't have a better answer.  Lets get this working while we thing more.

brownd1978 (approved):

PR #101, opened 2021-09-23
Mix campaign improvements

soleti:
This PR includes two main changes:

- `FlateMinus` and `NoPrimaryPBI` stages added to the POMS configuration
- Fixed `IPAStop` fcl files

### DIFF
--- a/JobConfig/primary/IPAStopParticle.fcl
+++ b/JobConfig/primary/IPAStopParticle.fcl
@@ -1,0 +1,52 @@
+#
+# Base configuration for Resampling IPA muon stops and generating primary particles from them
+#
+# original author: S Middleton
+#
+#include "Offline/fcl/standardServices.fcl"
+#include "Offline/EventGenerator/fcl/prolog.fcl"
+#include "Production/JobConfig/common/prolog.fcl"
+#include "Production/JobConfig/primary/prolog.fcl"
+#include "Production/JobConfig/pileup/prolog.fcl"
+# name all processes the same, to mask any provenance in ensemble creation
+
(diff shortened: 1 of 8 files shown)


JobConfig/primary/IPAStopParticle.fcl
```diff
@@ -0,0 +1,52 @@
+#
```
brownd1978, line 1: This is almost identical to TargetStopParticle. I suggest to create StopParticle.fcl with the common lines, and #include that in both TargetStopParticle and IPAStopParticle, then add the few lines that are different.

MDC2020/gen_Mix.sh
```diff
@@ -20,11 +20,11 @@ is necessary to provide, in order:
...
 - the campaign version of the input file [campaignInput],
 - the campaign version of the output file [campaignOutput],
-- the number of events per job [nEventsPerJob],
-- the number of jobs [nJobs].
+- the number of events per job [nEventsPerJob] (needed only for NoPrimary),
+- the number of jobs [nJobs] (needed only for NoPrimary).
 
 Example:
-    gen_Mix.sh CeEndpoint MDC2020 k m 1000 100
+    gen_Mix.sh CeEndpoint MDC2020 k m
```
brownd1978, line 28: thanks for updating the comment

JobConfig/primary/IPAStopParticle.fcl
soleti: Ok, I created `StopParticle.fcl`, I think it should be fine now

brownd1978 (approved):

brownd1978:
@FNALbuild run build test

sophiemiddleton (approved):
Looks good, yes I think the PDGId is for the mother now, I see you fixed it so should be fine

PR #102, opened 2021-09-29
Cleanup indentation and white space in production fcl, and remove obsolete files

brownd1978:
Also update gen_Reco to use sam.
Also remove obsolete MDC2018 scripts for generating primaries using tuple-form muon stops files.  There is no tuple-form MDC2020 stops, and the outputs are incompatible with MDC2020 downstream processing.  The modules which produce those primaries must be rewritten.
Also remove obsolete 'examples' directory, which contained scripts used to run generate_fcl in MDC2018.  The updated equivalent scripts are in the MDC2020 directory.
Also remove obsolete MDC2018 style ceMixDigi script.  This functionality was replaced months ago by ceMix.


brownd1978:
@FNALbuild run build test

brownd1978:
@FNALbuild run build test

brownd1978:
@FNALbuild run build test

kutschke:
@brownd1978 There is a conflict in JobConfig/beam/prolog.fcl .  Please resolve it.

I see that you have mixed in some real changes along with the changes to white space and comment characters.  I will make a list of those and decide if any need reviewing.

kutschke:
@brownd1978  Please edit the PR introductory comments to mention the major deletions and say a bit about them.

kutschke:
@soleti Please look in the conversation to find a post from me that calls out the non-trivial changes.  Please focus your review on these plus the added files.

kutschke:
@brownd1978 @HCasler 

Dave: do you understand why the most recent CI test (that reported at 2:01 AM CDT)  today is clean even though the PR says that there is a merge conflict?   Was the conflict created by a push after that test ran?  The commit times suggest no but the commit times are just that - it does not say when you pushed to your GitHub fork.

Helenka: please check what the CI does when there is a merge conflict.  I would have expected it to fail at the merge and stop.

JobConfig/beam/prolog.fcl
```diff
@@ -5,301 +5,301 @@ BEGIN_PROLOG
...
+  #
+  # define products that need to be kept
+  #
   SimKeptProducts : [
-            "keep art::EventIDs_*_*_*",
-	    "keep mu2e::GenParticles_*_*_*",
-            "keep mu2e::GenEventCount_*_*_*",
-            "keep mu2e::StatusG4_*_*_*" 
+    "keep art::EventIDs_*_*_*",
+    "keep mu2e::GenParticles_beamResampler_*_*",
```
soleti, line 290: this is expected right? we want to keep only the beamResampler GenParticles?

MDC2020/gen_OffSpill.sh
```diff
@@ -0,0 +1,24 @@
...
+#
+# this script requires mu2etools, mu2efiletools and dhtools be setup
+# It also requires the SimEfficiencies for the beam campaign be entered in the database
+# $1 is the name of the primary (ie CeEndpoint, etc).
+# $2 is the dataset description
+# $3 is the campaign version of the input file.
+# $4 is the campaign version of the output and primary file.
+# $5 is the number of input collections to merge (merge factor)
+rm template.fcl
+samweb list-file-locations --schema=root --defname="dts.mu2e.$1.$2$4.art"  | cut -f1 > $1$4.txt
```
soleti, line 11: shouldn't the dataset be `$2$3` (and the name of the txt file `$1$3.txt`)

MDC2020/gen_OffSpill.sh
```diff
@@ -0,0 +1,24 @@
...
+echo '#include "Production/JobConfig/digitize/OffSpill.fcl"' >> template.fcl
+echo outputs.TriggeredOutput.fileName: \"dig.owner.${1}OffSpillTriggered.version.sequencer.art\" >> template.fcl
+echo outputs.UntriggeredOutput.fileName: \"dig.owner.${1}OffSpillUntriggered.version.sequencer.art\" >> template.fcl
+generate_fcl --dsconf="$2$4" --dsowner=mu2e --description="$1OffSpill" --embed template.fcl \
+  --inputs="$1$4.txt" --merge-factor=$5
```
soleti, line 16: Here as well I think it's `$1$3`?

MDC2020/gen_OnSpill.sh
```diff
@@ -0,0 +1,24 @@
...
+#
+# this script requires mu2etools, mu2efiletools and dhtools be setup
+# It also requires the SimEfficiencies for the beam campaign be entered in the database
+# $1 is the name of the primary (ie CeEndpoint, etc).
+# $2 is the dataset description
+# $3 is the campaign version of the input file.
+# $4 is the campaign version of the output and primary file.
+# $5 is the number of input collections to merge (merge factor)
+rm template.fcl
+samweb list-file-locations --schema=root --defname="dts.mu2e.$1.$2$4.art"  | cut -f1 > $1$4.txt
```
soleti, line 11: Same, as above for the input I think it should be `$2$3`

MDC2020/gen_OnSpill.sh
```diff
@@ -0,0 +1,24 @@
...
+echo '#include "Production/JobConfig/digitize/OnSpill.fcl"' >> template.fcl
+echo outputs.TriggeredOutput.fileName: \"dig.owner.${1}OnSpillTriggered.version.sequencer.art\" >> template.fcl
+echo outputs.UntriggeredOutput.fileName: \"dig.owner.${1}OnSpillUntriggered.version.sequencer.art\" >> template.fcl
+generate_fcl --dsconf="$2$4" --dsowner=mu2e --description="$1OnSpill" --embed template.fcl \
+  --inputs="$1$4.txt" --merge-factor=$5
```
soleti, line 16: Same as in `gen_OffSpill`, the input is `$1$3.txt` I think

brownd1978:
This conflict appears spurious: the GitHub diff algorithm is choking on simultaneous changes to indenting and replacing "//" comments with "#".  I explicitly merged back main, 'resolved' the conflict, and recommitted, but the result was identical to what was already there.   Github now says 'no conflicts'.

brownd1978:
@FNALbuild run build test

soleti (commented):
There are few comments regarding the `gen_OnSpill.sh` and `gen_OffSpill.sh` scripts which I think contain the wrong parameter

gaponenko:
What is the motivation for replacing '//' with '#'? Both of them are valid 
ways to start a fhicl comment, but '//' is not ambiguous, while '#' is 
also used for '#include', so IMO '//' is preferable.

Andrei

MDC2020/gen_OffSpill.sh
brownd1978: In mixing we allow the primary to have a different version than the mixin files, but we assume the primary was produced with the same version as the current production.  We can make the same assumption here and then we need only 1 version variable.  Since primaries are cheap to produce I think it's reasonable.

brownd1978: Actually this logic might not be right: cosmics are not a cheap primary, and if we ever want to re-digitize a cosmic primary with a later version of Offline this script wouldn't work.  I'll just fix it according to your suggestion.

JobConfig/beam/prolog.fcl
brownd1978: Good catch.  This was intended to eliminate a duplicate copy of GenParticle that showed up after resampling (pre/post compression) but I see now that it would break the POT output.    I've reverted it.

brownd1978:
@FNALbuild run build test

kutschke:
Thanks for the improved title and description and for fixing the spurious conflict.

brownd1978:
// comments were changed to # to provide internal consistency (not mixed in 1 file), conformity with majority opinion, and for consistency with vim perl mode.

JobConfig/beam/prolog.fcl
soleti: I think this still hasn't been reverted?

brownd1978:
@FNALbuild run build test

soleti (approved):

PR #111, opened 2021-10-08
Split mixing campaign in one campaign for each primary

soleti:
- Adding the reco stage to mixed and unmixed primaries made the `mdc2020_mix` campaign too complicated. I split it into one campaign for each primary (at the moment `NoPrimary`, `CeEndpoint`, `CePlusEndpoint`, `FlateMinus`, `FlatePlus`). 
- Now the job loads the cfg file from `cvmfs` and not from my local area
- The mixing+digi+reco 1% campaigns will name the output dataset with the description `MDC2020km` since they use `k` tag as input and `m` tag as output.

### DIFF
--- a/Scripts/gen_Mix.sh
+++ b/Scripts/gen_Mix.sh
@@ -6,7 +6,7 @@
 # $2 is the dataset description
 # $3 is the campaign version of the input files for the mixing.
 # $4 is the campaign version of the primary files.
-# $5 is the campaign version of the output and primary file.
+# $5 is the campaign version of the output files.
 # $6 is the number of events per job. 
 # $7 is the number of jobs.
 
--- a/CampaignConfig/mdc2020_noprimary.ini
+++ b/CampaignConfig/mdc2020_noprimary.ini
@@ -15,19 +15,19 @@ param_overrides="[]"
 test_param_overrides="[]"
 merge_overrides=False
 login_setup=mu2epro_login
-job_type=mu2e_cvmfs_noprimary_production_jobtype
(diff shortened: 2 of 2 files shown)


Scripts/gen_Mix.sh
```diff
@@ -4,37 +4,39 @@
 # It also requires the SimEfficiencies for the beam campaign be entered in the database
 # $1 is the name of the primary (ie CeEndpoint, etc).
 # $2 is the dataset description
-# $3 is the campaign version of the input file.
-# $4 is the campaign version of the output and primary file.
-# $5 is the number of events per job. 
-# $6 is the number of jobs.
+# $3 is the campaign version of the input files for the mixing.
+# $4 is the campaign version of the primary files.
+# $5 is the campaign version of the output and primary file.
```
brownd1978, line 9: I think $5 is just the output,  right?

brownd1978 (commented):
There's a minor fix needed in the comment of gen_Mix.
I have a design question: the primary (mixing and digi) cfg and ini look the same as each other, which means a lot of code duplication and parallel maintenance when we change those.   Could those be consolidated?  ie could you use global.stage_name in place of (say) FlateMinus?  I guess this requires #include, which you explained is not available currently.

Scripts/gen_Mix.sh
soleti: Fixed

soleti:
`global.stage_name` is a good suggestion, although the name of the stage (e.g. `[stage_flateplusmix]`) will stay the same. But unfortunately there is no way to avoid the repeated code because there is no #include feature.

PR #113, opened 2021-10-08
Early mixing support

brownd1978:
This PR adds support for early (no time cut) mixing.  It isn't complete, as the digitzation window is still the normal one, but it uses the correct mixin collections with the correct scalings.
This PR also centralizes the Pileup prescale factors, so those are defined once/campaig, with the same values sourced for both pileup generation and mixing.


bonventre (approved):

brownd1978:
@FNALbuild run build test

JobConfig/mixing/epilog.fcl
```diff
@@ -4,12 +4,12 @@ physics.producers.EWMProducer.SpillType : 1
 services.GeometryService.inputFile      : "Offline/Mu2eG4/geom/geom_common_reco.txt"
 # add primary and mixers to the list of step producers.  Note the StrawDigi maker uses a general selector if no module is specified
 physics.producers.makeSD.StrawGasStepModule : ""
-# configure the mixin efficiences
+# configure the mixin efficencies
```
soleti, line 7: efficiencies, I guess?

Scripts/gen_EleBeamResampler.sh
```diff
@@ -1,23 +1,35 @@
...
-echo '#include "Production/JobConfig/beam/EleBeamResampler.fcl"' >> template.fcl
-echo physics.filters.beamResampler.mu2e.MaxEventsToSkip: ${nskip} >> template.fcl
+# write the elebeamresampler.fcl
+rm elebeamresampler.fcl
+echo '#include "Production/JobConfig/pileup/EleBeamResampler.fcl"' >> elebeamresampler.fcl
+echo physics.filters.beamResampler.mu2e.MaxEventsToSkip: ${nskip} >> elebeamresampler.fcl
 #
-generate_fcl --dsconf=MDC2020$1 --dsowner=brownd --run-number=1202 --description=EleBeamResampler --events-per-job=200000 --njobs=100 \
-  --embed template.fcl --auxinput=1:physics.filters.beamResampler.fileNames:EleBeamCat.txt
+generate_fcl --dsconf=$1$3 --dsowner=brownd --run-number=1202 --description=EleBeamResampler --events-per-job=400000 --njobs=$4 \
```
soleti, line 26: is there a reason why we doubled the events per job here?

brownd1978: I estimate these jobs will take the canonical 2 hours with 400K.   These scripts are intended for use in Validation, if you plan to use them also in the campaign we would have to update that.

PR #118, opened 2021-10-12
PBI script filenames

soleti:
The `CreatePBISequence.sh` now creates the files with the Mu2e naming convention.

### DIFF
--- a/Scripts/CreatePBISequence.sh
+++ b/Scripts/CreatePBISequence.sh
@@ -6,11 +6,13 @@
 # $2 is the PBI sequence type (either Normal or Pathological)
 # $3 is the maximum number of events/job
 # $4 is the run number
+# $5 is the user field for the filename
+# $6 is the description field for the filename
 #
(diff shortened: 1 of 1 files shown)


Scripts/CreatePBISequence.sh
```diff
@@ -6,11 +6,16 @@
...
+# $6 is the description field for the filename
 #
+# Example:
+# ./CreatePBISequence.sh 33344 Normal 1000 1201 mu2e MDC2020
+
 infile=PBI_$2_$1.txt
 nlines=`wc -l < $infile`
 let nfiles=$nlines/$3
-outroot=PBI_$2_$1_
+outroot=sim.$5.PBI$2_$1.$6.
```
brownd1978, line 18: Is the "." at the end of outroot intentional?

soleti: Yes, it's for the identifier field, otherwise it is just attached to the description field.

PR #124, opened 2021-10-20
Campaigns fix

soleti:
This PR includes:
- Local version of the job types for quick testing (not used by default)
- CORSIKA mixing campaign
- Bug fixes (path for mcs files)

### DIFF
--- a/JobConfig/cosmic/S1DSStops.fcl
+++ b/JobConfig/cosmic/S1DSStops.fcl
@@ -27,7 +27,9 @@ physics: {
     genCountLogger: { module_type: GenEventCountReader makeHistograms: false }
   }
 
-  PrimaryPath :  [@sequence::Common.generateSequence, @sequence::Common.g4Sequence, stepPointMomentumFilter, cosmicFilter, compressPV]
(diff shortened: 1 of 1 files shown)


JobConfig/mixing/CreatePBISequence.fcl
```diff
@@ -12,6 +12,7 @@ services: {
 
 source: {
   module_type :  PBISequence
+  reconstitutedModuleLabel : PBISim
```
brownd1978, line 15: I don't know why this shows up as a difference, this line is already merged.

brownd1978 (commented):
I believe this PR depends on other pending PRs in Production and Offline, so it shouldn't be merged yet.

JobConfig/mixing/CreatePBISequence.fcl
soleti: Yeah, I also noticed this, I am not sure

brownd1978:
I guess it doesn't formally depend on the other PRs, but to produce the output we want it needs changes to the OffSpill trigger and removal of the SignalFilter from mixing.

brownd1978:
@FNALbuild run build test

PR #130, opened 2021-10-28
POMS campaigns refactoring

soleti:
This PR includes a refactoring of the POMS configurations which now use the `includes` statement to avoid repeated settings.


brownd1978:
@FNALbuild run build test

PR #137, opened 2021-11-13
Beam 10% campaign

soleti:
POMS configuration updates for the beam 10% campaign.


brownd1978 (commented):
For beam production this is fine.
I'd like to confirm the mixing configuration for 10%.  We want to keep  the size of the pileup (mixin) files the same, but increases their numbers.  Can you verify that the concatenation stage does this?   For now we should keep the number of mixing jobs the same.  We might increase the # of mixing jobs after we re-evaluate the oversampling.

PR #152, opened 2022-01-14
Mu2eII_SM21: Merge head for event generators

sophiemiddleton:


kutschke (approved):
This was tested together with Offline PR #686 and it passed those tests.  

After an off-github discussion with Sophie, we decided that threshold for merging is that it passes the CI. Any additional testing is the responsibility of the Mu2e-II co-convenors.

PR #156, opened 2022-02-17
Use CrvRecoSequence

brownd1978:


ehrlich-uva:
Lines 121 to 128 can be removed in Validation/cosmicSimReco.fcl. See my comment at Offline PR711.

PR #157, opened 2022-02-21
Remove a wrong comment and tweak cuts for stopped muon validation job.

gaponenko:
Small tweaks to the stopped muon validation job.

### DIFF
--- a/JobConfig/validation/stoppedMuonsSingleStage.fcl
+++ b/JobConfig/validation/stoppedMuonsSingleStage.fcl
@@ -128,8 +128,10 @@ physics.producers.g4run.Mu2eG4CommonCut: {
 
 physics.producers.g4run.SDConfig.enableSD: [ virtualdetector ] # activate just the explicitly listed SDs
 
-# Not too coarse - we care about stopped muon positions.
-physics.producers.g4run.physics.minRangeCut : 0.010 # mm
+# Coarser cuts per e-mail discussion with Krzysztof Genser on 2022-02-18
+physics.producers.g4run.physics.minRangeCut : 1. # mm
+physics.producers.g4run.physics.protonProductionCut : 1. # mm
+physics.producers.g4run.physics.turnOnRadioactiveDecay : true
 
 services.GeometryService.inputFile : "Offline/Mu2eG4/geom/geom_common_current.txt"
 services.TFileService.fileName : "nts.owner.stoppedMuonsSingleStage.version.sequencer.root"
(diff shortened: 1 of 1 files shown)


JobConfig/validation/stoppedMuonsSingleStage.fcl
```diff
@@ -128,8 +128,10 @@ physics.producers.g4run.Mu2eG4CommonCut: {
 
 physics.producers.g4run.SDConfig.enableSD: [ virtualdetector ] # activate just the explicitly listed SDs
 
-# Not too coarse - we care about stopped muon positions.
-physics.producers.g4run.physics.minRangeCut : 0.010 # mm
+# Coarser cuts per e-mail discussion with Krzysztof Genser on 2022-02-18
+physics.producers.g4run.physics.minRangeCut : 1. # mm
+physics.producers.g4run.physics.protonProductionCut : 1. # mm
+physics.producers.g4run.physics.turnOnRadioactiveDecay : true
```
resnegfk, line 134: This should be turnOffRadioactiveDecay, not On, 
as the idea is that those muons do not come from the radioactive decays.
One should probably check if one would see a statistically significant difference in the number of stopped muons, while seeing CPU savings.

resnegfk (changes requested):

resnegfk:
The fact that I submitted a review and requested changes was somehow "reset" when Rob made me a reviewer again after Andrei has done it is strange.

kutschke:
Ooops.  Sorry about that. That surprises me too. 

I would also like Andrei to run a test to let us know what difference this makes.

gaponenko:
Krzysztof, thank you for catching this, the radioactive decay setting is 
fixed now.  Turning them off was what you originally suggested and I meant
to do, but then I started looking at the code (wondering why we had both
turnOnRadioactiveDecay and turnOffRadioactiveDecay), and ended up using
the wrong one.

As for testing:  a job with 1e6 events ran overnight (with decays on)
got 1517 stopped muons.  The nightly validation shows the stopping
rate of 1592e-6.   If the nightly job also uses 1e6 events, the
two numbers are within 1.3 sigma of each other.

I did not see a large CPU change with decays on, but it is reduced
after turning decays off:

original:  352 user s / 10e4 events
commit-1:  317 user s / 10e4 events  
commit-2:  278 user s / 10e4 events

Andrei

resnegfk (approved):
Thank you for addressing this.
The difference in the number of events may be due to event fluctuations.
The added cuts will save the CPU and should still leave the test sensitive enough for what we need it.

PR #161, opened 2022-04-01
Update mixing configuration to optimize mixin stream randomization with full production

brownd1978:


brownd1978:
I've tested this by generating and running NoPrimary using MDC2020p output, and running ceSimReco.

PR #169, opened 2022-04-14
Speeding up G4 cosmic resampling

soleti:
Upon @oksuzian suggestion I added few lines to `S2Resampler.fcl` which give a factor of ~2 speedup. However, I am not exactly sure of what these modifications are actually doing, @oksuzian could you please say a bit more? Thanks.

### DIFF
--- a/JobConfig/cosmic/S2Resampler.fcl
+++ b/JobConfig/cosmic/S2Resampler.fcl
@@ -41,11 +41,6 @@ physics.producers.g4run.inputs.simStageOverride: 1
 physics.producers.g4run.inputs.inputPhysVolumeMultiInfo: "CosmicResampler"
 physics.producers.g4run.inputs.updateEventLevelVolumeInfos.input: "CosmicResampler:eventlevel"
 physics.producers.g4run.inputs.updateEventLevelVolumeInfos.outInstance: "eventlevel"
-# Geant4 physics settings to speed up simulation
-physics.producers.g4run.physics.minRangeRegionCuts: { CalorimeterMother : 0.1 TrackerMother : 0.001 }
-physics.producers.g4run.physics.minRangeCut : 1.0 # mm
-physics.producers.g4run.physics.physicsListName: ShieldingM
-physics.producers.g4run.physics.useEmOption4InTracker: true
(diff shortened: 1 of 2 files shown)


oksuzian (commented):
We need to check that the impact on physics is minimal by introducing these cuts.
I will simulate a small CRY sample with and without these cuts to estimate the reconstruction efficiencies.

resnegfk (changes requested):
The placement of the g4run.physics changes does not seem to be correct,
if JobConfig/cosmic/S2Resampler.fcl is to be the main inpu fcl file,
as some of the params are overwritten in the Production/JobConfig/primary/epilog.fcl

resulting in:
            minRangeRegionCuts: {
               CalorimeterMother: 1e-1
            }
            noDecay: []
            physicsListName: "ShieldingM_EMZ"

Perhaps one should create another epilog for that stage?

Other than that the choice of the intended parameters seems to be reasonable,
if one want to have a detailed simulation in the tracker, but not elsewhere.
It would seem one should use SheildingM_EMZ in stages involving DS/tracker only.
The:
physics.producers.g4run.physics.physicsListName: ShieldingM
physics.producers.g4run.physics.useEmOption4InTracker: true
is close to that, and seem appropriate for cosmic simulations.

soleti:
Thank you for your comment @resnegfk, you're right those settings are overwritten by the epilog. @oksuzian can you make sure your test files have the correct settings? Once we confirm there are no significant changes in the outcome I will modify this PR and include a dedicated epilog.

soleti:
I moved the settings to a dedicated epilog. @oksuzian verified that the two versions (with and without these settings) agree at the 0.1% level.

oksuzian (approved):
These are the settings we've tested.
It improves the resampling stage by about 50% without a visible impact on physics.

resnegfk (approved):
Apart from the lack of the new line in JobConfig/cosmic/resampling_epilog.fcl
(would the style checker flag that? :) )
it seems fine now

brownd1978:
@FNALbuild run build test

brownd1978:
Please fix the missing newline in resampling_epilog.fcl

PR #168, opened 2022-04-13
Digitization stream split

brownd1978:
Split digitization stream into functional output files.  This also simplifies the NoPrimary mixing workflow by treating NoPrimary generation as a 'normal' primary.

PR #170, opened 2022-04-15
Adapt to separation of BField config from Geometry config

kutschke:
This is needed for Offline PR #765.  See also the conversation for that PR.

Find all references to geom_common_reco.txt and replace with
   services.GeometryService.inputFile      : "Offline/Mu2eG4/geom/geom_common.txt"
   services.GeometryService.bFieldFile     : "Offline/Mu2eG4/geom/bfgeom_reco_v01.txt"

When this process converges we will remove geom_common_reco.txt


kutschke:
These failures were expected.  This is being retested in Offline PR 765.

AndrewEdmonds11 (approved):
Looks good to me!

soleti (approved):

brownd1978:
Let me know if its not clear how to merge the changes from 168 into this PR.

kutschke:
> Let me know if its not clear how to merge the changes from 168 into this PR.

It looks straightforward.  I am looking at it now.

kutschke:
The conflicts created by the merge of #168 have been fixed.   That merge also created a new point of maintenance, JobConfig/digitize/epilog.fcl ; that's fixed too.

@FNALbuild run build test with Mu2e/Offline#765

rlcee (approved):
It is a little disconcerting to see a geom file replaced by a bfield file, but I think this has the intended effect due to the nature of the geom file, and I'm sure you have checked this point.

kutschke:
> It is a little disconcerting to see a geom file replaced by a bfield file, but I think this has the intended effect due to the nature of the geom file, and I'm sure you have checked this point.

It looked weird to me at first as well.

In cases where the geometry file override existed only to override the BField, the default geometry file (geom_common) is correct and does not need to be overridden.  In some of the changes added today the geometry file override changes both geometry and bfield - so both are present in the new version.  See, for example, JobConfig/common/epilog_muplus_DS50_reco.fcl

kutschke:
@FNALbuild run build test with Mu2e/Offline#765

PR #171, opened 2022-04-18
Add editor backup files to the .gitignore.

kutschke:


.gitignore
```diff
@@ -1,6 +1,10 @@
 # Prerequisites
 *.d
 
+# Editor backups
+*~
+\#*
```
gaponenko, line 6: Hi Rob, is this '#*' pattern meant for emacs files?  Two comments:  (1) the names I see start with a dot  '.#*',   (2) they are not backup files but an indication that the file has not been saved after an edit, so what you see on the screen is not what git would commit from the disk.   I don't think it is useful to ignore those.  One should sync editor buffers to disk before a commit.

brownd1978 (approved):

PR #174, opened 2022-04-26
Updating digi and mixing scripts

soleti:
Updates about database purposes and versions. Also fixed documentation.


Scripts/gen_Digitize.sh
```diff
@@ -8,32 +8,37 @@
...
+# $8 is the database version
 primary=$1
 name=$primary.$2$3
 conf=$2$4
 merge=$5
 digitype=$6
 digout=${primary}${digitype}
-dbpurpose=$2_DIGI
-dbver=$7
+dbpurpose=$7
```
brownd1978, line 19: I think we want $2_$7, like in gen_Mix

brownd1978 (changes requested):

brownd1978 (approved):

PR #176, opened 2022-04-27
Save only output version in description of mixed outputs

soleti:


brownd1978 (approved):

brownd1978:
This makes the output of Mix and regular digitization have the same format, which is nice.

PR #183, opened 2022-05-17
fixed gen_mix for noprimary

soleti:
This PR contains two fixes: 
- When the beam model is "Seq" it includes the right mixing FCL file
- The condition ` [ "${primary}" == "NoPrimary*" ]` doesn't work. The right one would be `[[ ${primary} == NoPrimary* ]`. However, since now the PBI case is managed by the "Seq" if, we don't need the * wildcard anymore.


brownd1978 (approved):

brownd1978:
@FNALbuild run build test

PR #190, opened 2022-06-12
Cosmicsplit

brownd1978:
script to pre-select Cosmic events prior to digitization.  Currently this selects signal-like tracks, but in future it should select OnSpill and OffSpill calibration events too, as separate streams.

### DIFF
--- a/JobConfig/cosmic/SpillSplitter.fcl
+++ b/JobConfig/cosmic/SpillSplitter.fcl
@@ -1,6 +1,6 @@
 #
-# split cosmic primarys prior to OnSpill or OffSpill digitization.  OnSpill (mixing)
-# requires there be a signal-like particle, Offspill is simply prescaled
+# select cosmic primarys for digitization.  OnSpill (mixing) requires a signal
+# or calibration-like track, # Offspill is simply prescaled
 #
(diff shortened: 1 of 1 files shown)


kutschke:
@oksuzian  and @soleti  : a gentle reminder that this PR is in your input queue.

soleti (approved):

kutschke:
@FNALbuild run build test

oksuzian (approved):
I think we can relax some of the parameters:
MinimumTrkPlanes : 0 // Not sure why it's zero
MinimumTrkPlaneSpan : 16 // can lower it
MaxImpact : 150.0 // can make it higher
However, that's a detail, it can be optimized in the future.

brownd1978:
Why would you loosen the impact cut?  That will just increase the acceptance of non-signal-like particles.  Based on the digi study I did it could be tightened, maybe to 125mm.   The nplanes cut was doing very little, so I just turned it off.  The plane span cut could be adjusted, but I'm not sure what you think needs to be improved; rejection of non-signal-like particles or acceptance of signal-like?

kutschke:
FYI:  I have already merged it, since it was approved by both reviewers - you will need a new PR if you want to make changes.

PR #191, opened 2022-06-14
Fixing RPC scripts

soleti:
We fix the `gen_Primary.sh` script and the RPC fcl files to generate and mix RPC primaries.


brownd1978 (approved):

sophiemiddleton (approved):
looks good

PR #192, opened 2022-06-16
set of FHICL files for producing dummy samples for characterizing/improving the pattern recognition in the pi+ calibration special run

gianipez:
All the following processes are studied with the B-field at 0.7 T
- the π+ -> e+ nu process at rest in the Stopping Target (ST)
- π+ and µ+ with p>50MeV/c in the ST (to avoid them being stopped and let them decay in flight)
- flat e+


soleti:
Thanks @gianipez, a general comment: instead of copy-pasting the content of `FlatePlus.fcl`, could you maybe include it and add only the configurations that need to be changed?

gianipez:
Hi folks, 
Are there any other changes required before approving this PR?

JobConfig/primary/PiPlusCalib_bfgeom_DS70.fcl
```diff
@@ -0,0 +1,15 @@
+#
+# generate and produce Detector Steps from flat spectrum using mu+ target stops
+#
+# original author: S Middleton
```
soleti, line 4: Maybe you could change the description here to be a bit more complete and change the author

brownd1978 (changes requested):
Hi Giani, The way we have been dealing with this situation recently is different: instead of creating every possible combination of BField, geometry, process etc as an explicit fcl file (which will lead to exponential expansion and confusion), we instead have scripts that generate the template fcl for a given set on-the-fly, given command line arguments.  This is the only practical way to deal with the fact the fhicl isn't a language (doesn't have variables or logical statements).  I'd like you to consider doing the same for these scripts.  Please look at Production/Scripts/gen_Primary.fcl for an example.

soleti:
`generate_fcl` is interpreting `-11` as an argument I think because of the minus sign. Instead of feeding the parameter as auxiliary input can't you add it to the `primary.fcl` file? That should work.

Scripts/gen_FlatMuDaugther.sh
```diff
@@ -0,0 +1,78 @@
...
+# $5 is the kind of input stops (Muminus, Muplus, IPAMuminus, IPAMuplus, or Cosmic)
+# $6 is the number of jobs
+# $7 is the number of events/job
+# $8 is the pdgId of the paritcle to generate
+# $9 is the startMom
+# $10 is the endMom
+# $11 is the name of the BField file
+#
+# example for running this script:
+#        source Production/Scripts/gen_FlatMuDaugther.sh FlatePlus MDC2020 k v01 Muminus 2 100 -11 30 100 "Offline/Mu2eG4/geom/bfgeom_DS70_no_tsu_ps_v01.txt"
```
soleti, line 23: Daugther is misspelled here and in the filename

soleti (changes requested):

Scripts/gen_FlatMuDaugther.sh
soleti: Also, can you replace `v01` with our convention, for example `s`? particle is also misspelled

Scripts/gen_FlatMuDaughter.sh
```diff
@@ -14,13 +14,13 @@
...
 # $7 is the number of events/job
-# $8 is the pdgId of the paritcle to generate
+# $8 is the pdgId of the particle to generate
 # $9 is the startMom
 # $10 is the endMom
 # $11 is the name of the BField file
 #
 # example for running this script:
-#        source Production/Scripts/gen_FlatMuDaugther.sh FlatePlus MDC2020 k v01 Muminus 2 100 -11 30 100 "Offline/Mu2eG4/geom/bfgeom_DS70_no_tsu_ps_v01.txt"
+#        source Production/Scripts/gen_FlatMuDaugther.sh FlatePlus MDC2020 k s Muminus 2 100 -11 30 100 "Offline/Mu2eG4/geom/bfgeom_DS70_no_tsu_ps_v01.txt"
```
soleti, line 23: still Daugther here :)

JobConfig/primary/PiPlusCalib_bfgeom_DS70.fcl
gianipez: done

soleti (approved):

Scripts/gen_FlatMuDaughter.sh
```diff
@@ -0,0 +1,87 @@
...
+fi
+
+samweb list-file-locations --schema=root --defname="$dataset"  | cut -f1 > Stops.txt
+# calucate the max skip from the dataset
+nfiles=`samCountFiles.sh $dataset`
+nevts=`samCountEvents.sh $dataset`
+let nskip=nevts/nfiles
+# write the template
+rm primary.fcl
+if [[ "${stype}" == "Cosmic" ]]; then
```
brownd1978, line 63: Does this script make sense for cosmics?  I guess not, since those 'stops' are very different from target or IPA stops.  I suggest to remove this option.

Scripts/gen_FlatMuDaughter.sh
```diff
@@ -0,0 +1,87 @@
...
+  echo "#include \"Production/JobConfig/cosmic/S2Resampler${primary}.fcl\"" >> primary.fcl
+else
+  echo "#include \"Production/JobConfig/primary/${primary}.fcl\"" >> primary.fcl
+fi
+echo physics.filters.${resampler}.mu2e.MaxEventsToSkip: ${nskip} >> primary.fcl
+echo physics.filters.${resampler}.mu2e.MaxEventsToSkip: ${nskip} >> primary.fcl
+echo physics.producers.generate.pdgId: ${pdg}            >> primary.fcl
```
brownd1978, line 70: Since this code can be any particle (not just muon) I suggest to rename the script 'gen_FlatStopDaughter.sh.

brownd1978 (changes requested):

Scripts/gen_FlatMuDaughter.sh
gianipez: My motivation was that it is using the module `FlatMuDaughter` to generate the particle (it's the one used in FlaeMinus/Plus.fcl). 
I can change it, but I thought it would have been good to have that explicit.

gianipez: Hello @brownd1978 , what is the last decision for this at the end?

brownd1978: OK.  We can adjust names if needed in future.

brownd1978 (approved):

PR #193, opened 2022-06-28
Split cosmic track reco and diag ttree creation

bonventre:


rlcee:
My test run output is in
/mu2e/data/users/rlc/test*
I don't feel strongly about these since it is your data, but here are my observations.
- I see you are filtering the reco output. I think the most standard procedure would be to not filter the reco output, then move the filter to the diag fcl path.  In my mind, once reco is run, it is never thrown away.  The other option would be write the full output to one file and a skim to another.
- drop mu2e::ProtonBunchTimeMC ?
- why is triggerinfo present when you haven't run the trigger?
- in reco you have declared several analyzer modules, but don't use them?
- in diag trackT, ntrack=0?
- in diag, I double-click on leaves like a0 and I see the bulk of useful output is around 100, but there are values to 10e6.  Are these outliers meaningful or failures or flag values?
- I wonder if a0, b0, a1, b1 might have more mnemonic names

bonventre:
Thanks for the comments - I've removed some unused modules and the filter for now. I think the filter was adding triggerinfo so that should go away. We'll keep ProtonBunchTimeMC since some analyzers may still need it and its not actually MC info here.

For the diag ttrees, ntrack should probably be renamed itrack, but =0 is expected for successful track fits. And a0 etc. can have very high values for horizontal tracks so these are actually normal successful fits.

rlcee:
> Thanks for the comments - I've removed some unused modules and the filter for now. I think the filter was adding triggerinfo so that should go away. We'll keep ProtonBunchTimeMC since some analyzers may still need it and its not actually MC info here.
> 
> For the diag ttrees, ntrack should probably be renamed itrack, but =0 is expected for successful track fits. And a0 etc. can have very high values for horizontal tracks so these are actually normal successful fits.

Looks OK, last check - did you want to add the track filter to the diag-maker?

bonventre:
I think if we're leaving it out of the reco script lets just leave it out completely for now

rlcee:
OK, I am not an official reviewer, but I "approve".  Let me know what runs you want processed.  Also I know Dave is about to make a new tag of Offline.  Maybe this PR can get tagged and included in a new SimJob so we can use that instead of just a commit..

rlcee (approved):

PR #195, opened 2022-07-07
Added MuBeam splitter FCL file

soleti:
This FCL file splits the muon beam output into mu+ and mu-.

### DIFF
--- a/JobConfig/beam/MuBeamSplitter.fcl
+++ b/JobConfig/beam/MuBeamSplitter.fcl
@@ -1,6 +1,6 @@
 #
-# split the Beam stream into muon and 'electrons' (everything else, but mostly electrons)
-# This is run during POT job output concatenation
+# split the Muon Beam stream into muons and antimuons
+# This is run after the beam stage
 #
 #include "Offline/fcl/minimalMessageService.fcl"
 process_name: mubeamsplitter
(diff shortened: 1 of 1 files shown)


JobConfig/beam/MuBeamSplitter.fcl
```diff
@@ -0,0 +1,24 @@
...
+  muminusout: {
+    module_type: RootOutput
+    fileName : "sim.owner.MuMinusBeamCat.version.sequencer.art"
+    outputCommands:   [ "keep *_*_*_*" ]
+    RejectEvents: [ selectMuons ]
+  }
+}
+
+physics.filters.muonSelector.ParticleCodes : [ [-13, "uninitialized" , "mu2eKillerVolume" ] ]
+physics.outpath: [ muplusout, muminusout ]
```
brownd1978, line 24: Can you exclude the muminusout from the path, since we don't need that?

soleti: We don't need it now but maybe it can be useful in the future? If this is about the production I can just not copy back the muminus files

soleti: Ok I removed it

brownd1978:
@FNALbuild run build test

brownd1978 (approved):

PR #196, opened 2022-07-15
Alternate BField

brownd1978:
allow alternate bfield to be specified in primary generation and digitization.


brownd1978:
Note that we need to add something to the output collection configuration field to specify the non-standard bfield.  I will update after your feedback.

brownd1978:
@FNALbuild run build test with Mu2e/Offline#828

sophiemiddleton (approved):

Scripts/gen_Digitize.sh
```diff
@@ -18,7 +19,11 @@ digitype=$6
 digout=${primary}${digitype}
 dbpurpose=$2_$7
 dbver=$8
-#
+#bfield="Offline/Mu2eG4/geom/bfgeom_no_tsu_ps_v01.txt"
+if [[ $# -eq 9 ]]; then
+  bfield="Offline/Mu2eG4/geom/$9"
```
rlcee, line 24: Might be a little more standard to not assume which directory the file is in and require the full path relative to the standard data path.  If you choose to not make this change, the relevant comment at the top of the file could make this choice more clear.

Scripts/gen_Digitize.sh
```diff
@@ -62,7 +67,9 @@ else
 fi
 echo "Generating digitization scripts for $primary conf $conf output $digout database purpose, version $dbpurpose, $dbver"
 
-rm digitize.fcl
+if [[ -f "digitize.fcl" ]]; then
```
rlcee, line 70: I think you would get the same result with one line
`rm -f digitize.fcl`

rlcee (approved):
I included two small comments, up to you to decide what to do, approving now

Scripts/gen_Digitize.sh
brownd1978: I'd prefer to leave this.  We've adopted that directory as the official source of BField maps, someone can add whatever maps they want to their clone of Offline to test, so it's not a restriction.  The command line for gen_*.sh are already too long/complicated.

Scripts/gen_Digitize.sh
brownd1978: good point.  I updated this script and all the rest for the same issue.

kutschke:
CI was run earlier this evening when Offline PR #828 was tested in combination with this one.  Merging now.

PR #197, opened 2022-07-26
Adding the option of having custom owners

soleti:
This PR allows non-production users to run digi and reco.

### DIFF
--- a/Scripts/gen_Digitize.sh
+++ b/Scripts/gen_Digitize.sh
@@ -23,7 +23,8 @@ dbver=$8
 #
 usage() { echo "Usage:
   source Production/Scripts/gen_Digitize.sh [primaryName] [datasetDescription] \
-    [campaignInput] [campaignOutput] [nmerge] [digitype] [dbpurpose] [dbversion]
+    [campaignInput] [campaignOutput] [nmerge] [digitype] [dbpurpose] [dbversion] \
+    [dsowner]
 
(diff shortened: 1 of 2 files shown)


brownd1978 (approved):
This looks fine.

gaponenko:
This effectively makes users type their usernames to define a non-production "owner".   It is prone to typos.   An alternative approach (used e.g. the mu2efiletools package) is to take owner from the whoami output

    --std                      An abbreviation for --owner=mu2e
    --my                       An abbreviation for --owner=`whoami`

soleti:
@gaponenko thank you for the suggestion. I agree hardcoding the owner is prone to typos, but the problem is that this command runs also on grid nodes. I don't know what is `whoami` in that case.

brownd1978:
Is it run as part of the POMS campaign?  Does POMS have a way of identifying the submitter?
There is also a non-POMS use case for this script, so we need a solution that works for both.

gaponenko:
Yes, on the grid node whoami is not useful.  But where does the original 
input to the script comes from - somehow a user must be able to input the 
"owner" interactively?  May be whoami can be used at that point and the 
value propagated to the nodes.

Andrei

soleti:
POMS indeed knows the submitter, in fact in the configuration I have:

```
executable.name = gen_Reco.sh
...
executable.arg_8 = %(submitter)s
```

so the only occasion when the user can put a wrong name is when the script is used interactively I guess.

PR #204, opened 2022-09-23
Fcl file for PlotSTMWaveforms

AndrewEdmonds11:
This PR includes the fcl file to run the PlotSTMWaveforms module introduced in Mu2e/Offline#856

### DIFF
--- a/JobConfig/stm/plotSTMWaveforms.fcl
+++ b/JobConfig/stm/plotSTMWaveforms.fcl
@@ -1,0 +1,35 @@
+#
+# Analyze STMHits and create histogram of energy spectrum
+#
+
+#include "Offline/fcl/standardServices.fcl"
+
+process_name: PlotSTMSpectrum
+
+source : {
+  module_type : RootInput
+}
+
+services : @local::Services.Reco
+physics: {
+  producers : {
+  }
+  filters : {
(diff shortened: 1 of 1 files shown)


brownd1978 (commented):
The only analyzers run in Production/JobConfig currently are the ones that check the content of newly-produced data products (ie the compression checkers).  This looks like a standalone monitoring job, specific to STM.  Is there a reason you don't want it in Offline?  Alternatively, you could put it in Validation, or maybe we need a new directory 'Monitoring'.

rlcee:
> The only analyzers run in Production/JobConfig currently are the ones that check the content of newly-produced data products (ie the compression checkers). This looks like a standalone monitoring job, specific to STM. Is there a reason you don't want it in Offline? Alternatively, you could put it in Validation, or maybe we need a new directory 'Monitoring'.

To be more consistent with current trends, I'd put this fcl in STMReco/fcl if it is generally useful to the collaboration, and in STMReco/test if it is expert -only.

AndrewEdmonds11:
There was no particlar reason I put this fcl file in Production/JobConfig. I just thought we were trying to keep fcl files out of Offline. I will move it to Offline/STMReco and close this PR

PR #210, opened 2022-10-28
follow pdgid string in Offline

rlcee:
follows Mu2e/Offline#878  Probably should be merged after that


kutschke (approved):

kutschke (approved):
Offline 878 is merged.  Merging this one now.

Issue #134, opened 2021-11-04
Residual #include issues to fix


kutschke:
I did the following to find #include directives in Production do not start with either Offline/ or Production/ .  I presume we have no test coverage for these.  We should decide if the files are obsolete, and delete them, or if they should stay, and fix them.

I did the following from the head of Production:

find . -name \*.fcl -exec grep -H \#include {} \; | fgrep -v "Offline/"  | fgrep -v "Production/"

The output is:

./JobConfig/ensemble/reco-mcdigis-trig.fcl:##include "JobConfig/ensemble/prolog_reco.fcl"
./JobConfig/ensemble/reco-mcdigis-trig.fcl:##include "JobConfig/ensemble/epilog_reco.fcl"
./JobConfig/pbar/resample_vertex_s2.fcl:#include "minimalMessageService.fcl"
./JobConfig/pbar/resample_vertex_s2.fcl:#include "standardProducers.fcl"
./JobConfig/pbar/resample_vertex_s2.fcl:#include "standardServices.fcl"
./JobConfig/common/artcat.fcl:// for generate_fcl should #include this code, than have  a line that
./JobConfig/common/artcat.fcl://     #include "JobConfig/common/artcat.fcl"
./Validation/ceSimReco.fcl:##include "gen/fcl/Trigger/OnSpillTrigMenu/OnSpillTrigMenu.fcl"

rlcee:
It looks like most are commented out.  The pbar ones are not, but those show that they are even older.  I think these are from Bob B.'s second round of pbar studies.  I think they are still the most efficient existing  method to generate pbars, but so much fcl has changed recently, I may be wrong on that.  I'm happy to clean up any of these, just let me know

sophiemiddleton:
Should this issue be closed?

rlcee:
completed in #223

PR #223, opened 2022-12-22
fix small residual fcl paths

rlcee:
* fully addresses Mu2e/Production#134
* all changes are commented or fcl is obsolete, so no changes to Validation or production


kutschke (approved):
We discussed this before Ray did the work.

PR #224, opened 2023-01-19
Fifeutil updated to v3_6_0 in all .ini

sophiemiddleton:


kutschke (commented):
@sophiemiddleton  Do you need this merged quickly?

I see that there are still files in your personal disk space.   I understand that there are expediencies that are needed for development - is there a long term plan to get rid of files in the production manager's personal disk space?

I also see that you had to make the same change in many, many places.   Is there a plan to reduce multiple points of maintenance or is this a hard limitation of POMS?

sophiemiddleton:
Yes, the moving out of my area is the next task once NERSC is working. I dont want to introduce new places for errors until then. It's not urgent, unless anyone else is using these scripts.

sophiemiddleton:
As for the multiple places. This could be removed, this is just how it was set up. In reality we dont need new login defs for each process

kutschke:
> Yes, the moving out of my area is the next task once NERSC is working. I dont want to introduce new places for errors until then. It's not urgent, unless anyone else is using these scripts.

That time frame works for me.

kutschke:
> As for the multiple places. This could be removed, this is just how it was set up. In reality we dont need new login defs for each process

I don't know POMS well enough to give concrete advice.  As you move forward look for ways to minimize multiple points of maintenance.

kutschke (approved):

PR #230, opened 2023-02-05
add nightly validation fcl

rlcee:
This commit adds the fcl that we are running in the nightly validation jobs.  It was being kept with the nightly scripts or being partially generated in the jobs.  Putting it here satisfies a long-standing request from Dave.  Once this is merged, I'll adjust the scripts to use it, then Dave can start upgrading this fcl to more modern input files.  When adjusting this fcl, we want to keep the jobs under 1h on the central machines and keep one output file per job.  Also note that pileup is a combination of 4 variations (0-8,9,10-18,19) and it looks like ceMix is a set of jobs 0-9, duplicated to 10-19 with different seeds.  I don't recall the details, but could probably look up the emails.


kutschke (approved):
This is a pure addition of code, with no deletions or modifications.

PR #231, opened 2023-02-15
Script updates

sophiemiddleton:
Some new scripts examples of use of GetOpts for more user control


kutschke (commented):
I am curious about the introduction of the Old subdirectory.  Are you doing this just for archival purposes or do you expect that you might need to use these again? 

For example is generate_POT.sh the successor to Old/gen_POT.sh?

brownd1978 (commented):
If the gen_Digitize is no longer needed it should be removed, not just put in Old, as git will remember it.  If it is still needed it shouldn't be moved to Old.  Was the change from Digitize to Digitization intentional?

sophiemiddleton:
Old is just there because I'm still testing things, it will be removed once all the .cfg scripts are updated to use the new scripts. I just wanted review on these first before moving to the full transition, also I cant do that until we get the deg field study finished

kutschke:
> Old is just there because I'm still testing things, it will be removed once all the .cfg scripts are updated to use the new scripts. I just wanted review on these first before moving to the full transition, also I cant do that until we get the deg field study finished

Thanks - I will approve it now. Be sure to remember to remove it when you are done.

kutschke (approved):

brownd1978 (approved):

PR #237, opened 2023-03-10
updated scripts and cfg

sophiemiddleton:
Note this is an intermediate step, mdc2020_primary.cfg paths will return to CVFMS once test are finished


brownd1978:
@FNALbuild run build test

brownd1978 (approved):

PR #240, opened 2023-04-04
Remove time cuts for trigger hit reco in offspill/extracted

bonventre:


brownd1978 (commented):
I suggest epilog_OffSpill.fcl instead

brownd1978 (approved):
The names are consistent with vst usage.

PR #243, opened 2023-04-14
add reco validation files

rlcee:
- rm extracted_06, which was a spurious file
- add files to the reco jobs so that they are all a little over 1h

Only reco nightly validation should change.


kutschke (approved):
This looks OK to me.  I will leave the final decision to Dave since he is more in the loop than I am .

PR #252, opened 2023-05-08
Recent updates of cfg anf ini files which are known to work with current .sh scripts

sophiemiddleton:
These are changes which allow use of new scripts in the cfg files. Some further changes are needed


sophiemiddleton:
Hi Rob,
Yes, there's also unrelated changes such as those for CORSIKA and NoPrimary.

I havent done a test of the CRY scripts yet, I expect updates will be needed. I'm wiaitng to hear back from Dave to see if he ran tests.

sophiemiddleton:
OK, I have found out that I need to write the CRY ini/cfg so I will do that and push here once tested.

sophiemiddleton:
I have written the CRY scripts, they are untested but I think it would be worth moving forward with what we have since it would be good to get the new .fcl files in a musing to make testing easier

kutschke:
@brownd1978 Do you want to review this PR more or should we proceed with a Musing in order to facilitate addktional testing?

kutschke:
@FNALbuild run build test

brownd1978 (approved):
There is the usual chicken-egg recursion problems with musing, and I'm not really able to evaluate the POMS scripts.  We'll follow your suggestion and create the new build to finish this work.

PR #253, opened 2023-05-09
Fix reco

brownd1978:
I added in all the producers needed for production.  I'm not sure how this ever worked before.


kutschke (approved):
With this changed merged in to a test clone, I ran 20 events of Production/Validation/reco.fcl and the job completed normally.

rlcee (approved):

PR #254, opened 2023-05-11
A PR to contain small fixes after test, plus changes of paths to new musing

sophiemiddleton:


CampaignConfig/mdc2020_cry.cfg
```diff
@@ -1,9 +1,9 @@
 [global]
-includes = /mu2e/app/users/sophie/ProductionPush/Production/CampaignConfig/mdc2020_prolog.cfg
-desc = MDC2020v
+includes = /cvmfs/mu2e.opensciencegrid.org/Musings/SimJob/MDC2020x/Production/CampaignConfig/mdc2020_prolog.cfg
+desc = MDC2020x
```
kutschke, line 3: Is it time to look ahead and say that this will be MDC2020y ?

Until know I had not appreciated how many places haded to be fixed by hand to make this work.  I knew it was a lot but I had not seen i so clearly until now.   Does it make sense to bring this to the POMS team and ask they can see new POMS feature that would simplify this use case?

kutschke:
No sure why the CI did not run.  Trying it now.

@FNALbuild run build test

CampaignConfig/mdc2020_cry.cfg
sophiemiddleton: yes it would be great is there were just "version" parameter in the prolog or something and that we could have a %(version)s in the paths, but I dont think that works in the CFG paths

sophiemiddleton: and yes, I guess this will need to be MDC2020y, once I have all the pieces, I'll change it

kutschke:
CI worked for Offline PR980 a few minutes ago.  Let's see if the issue has passed.

@FNALbuild run build test

kutschke:
@sophiemiddleton Are you planning more commits to this PR?  If not, please make a commit commit with a trivial change to a comment so that I can test that the CI has recovered from it's earlier woes.  If you are planning more commits, I will wait for them.

kutschke (approved):

sophiemiddleton:
Hi Rob, I am still doing some work on the scripts, I will need to do some more checks to make sure they all work. Will leave this open and push here

sophiemiddleton:
I think as far as I can test these are good as an update, I expect further work to follow but I think that will be a separate PR

brownd1978 (approved):

brownd1978:
None of these files are in the CI coverage so there's no point running that

PR #265, opened 2023-09-06
ComboHitCollection Migration

brownd1978:
Drop trigger products in reco.  Remove references to StrawHitFlagCollection


kutschke (approved):
Tested with Offline PR 1066 and passed the CI tests.

PR #266, opened 2023-09-15
Script updates

sophiemiddleton:
Some on-going updates to the Production scripts. Includes:

* DIOTail campaign defintion
*  Fixes to use fife_util version which doesnt give python version error
* CORSIKA camapign for FNAL and NERSC are both defined
* Some changes to primary for reco dbversion differing from digi version (for reprocessing)


brownd1978 (commented):
Some specific questions in the PR.  I don't see any campaigns for reco reprocessing, did I miss it?

brownd1978 (approved):

PR #279, opened 2023-11-10
particle stops in the ST supporting wires

sdifalco:
add stuff to manage particle stops in the ST supporting wires

### DIFF
--- a/JobConfig/pileup/prolog.fcl
+++ b/JobConfig/pileup/prolog.fcl
@@ -301,8 +301,8 @@ Pileup: {
     "keep mu2e::CaloShowerSteps_compressDetStepMCs_*_*",
     "keep mu2e::CrvSteps_compressDetStepMCs_*_*",
     "keep mu2e::SimParticlemv_compressDetStepMCs_*_*",
-    "keep mu2e::GenParticles_compressDetStepMCs_*_*",
-    "keep mu2e::StepPointMCs_*_virtualdetector_*" ]
+    "keep mu2e::GenParticles_compressDetStepMCs_*_*"]
+#    "keep mu2e::StepPointMCs_*_virtualdetector_*" ]
   # define some common sequences
   beamResamplerSequence : [genCounter, protonTimeOffset, beamResampler]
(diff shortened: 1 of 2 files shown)


kutschke:
@sdifalco Does this PR need to be tested together with Offline 1126 (once that builds correctly?)

sdifalco:
> @sdifalco Does this PR need to be tested together with Offline 1126 (once that builds correctly?)

yes, this should work with Offline PR 1126

JobConfig/pileup/prolog.fcl
```diff
@@ -280,7 +301,8 @@ Pileup: {
     "keep mu2e::CaloShowerSteps_compressDetStepMCs_*_*",
     "keep mu2e::CrvSteps_compressDetStepMCs_*_*",
     "keep mu2e::SimParticlemv_compressDetStepMCs_*_*",
-    "keep mu2e::GenParticles_compressDetStepMCs_*_*" ]
+    "keep mu2e::GenParticles_compressDetStepMCs_*_*",
+    "keep mu2e::StepPointMCs_*_virtualdetector_*" ]
```
sophiemiddleton, line 305: might be best to comment this out for official produciton

sdifalco: yes I agree this would enlarge too much the dataset size, by the way it would be nice to have the possibility to compare for both e+ and e-, the track momentum at tracker entrance obtained by KinKal surface intersection with the particle momentum obtained by the virtual detector located at the tracker entrance.

kutschke: My preference is that the default be correct for production.  If people want to modify it for their development work or other studies, they are free to do so.  There are ways to accomplish this that do not involve commenting things out.   I have some ideas if you want to talk about them,

sdifalco: ok, just take this change away. We will find the way to insert the virtual detector information for generated e+e- in the gamma conversion  campaign if we consider it necessary.

sophiemiddleton (approved):

PR #280, opened 2023-11-12
Script updates

sophiemiddleton:
* should be changes for DIO tail and corsika jobs


brownd1978 (commented):
Some of the changes listed are old commits.   I suggest you merge the head of main to your branch, then push.  That should remove most of these.

brownd1978:
OK I see, your branch probably started from a very old commit, hence all the repeated, already merged commits.
Please merge the current head into your branch, then make a diff against the current head of Production so we can see the actual changes.  You will need to do that using Git, I don't think GitHub provides that as a tool.

brownd1978 (approved):

PR #282, opened 2023-11-15
Remerge

brownd1978:
This PR contains the commits from the script_update branch.  Please verify these are the changes you expect.

Those commits were PR 280, which was merged incorrectly to create the script_update branch, instead of being merged into main.


JobConfig/mixing/NoPrimary_pileup.fcl
```diff
@@ -0,0 +1,11 @@
+#
```
brownd1978, line 1: What is this file for?  It seems to be a fragment of the NoPrimary primary job.

sophiemiddleton (approved):

PR #286, opened 2023-11-29
start trigger validation

rlcee:
- trigger.fcl is not run-able, only nightly/trigger_0*.fcl

Some things to work on, doesn't have to be now, and might be easier with this already running:
- final decision on 
```
    SelectEvents : [ "*_highP_*_trigger" ]  (what is in the fcl now)
vs
    SelectEvents : [ "*_*_*_trigger" ]
```
- some KalSeed products might be missing
- it is not clear if the track TriggerInfo is correct
- can KalSeed histograms be made for only the seeds pointed to by TriggerInfo Ptr's?  Probably the easiest is to make a collection of these tracks, then the plots are automatic.  Making a validation module do this would probably require some re-working of the validation design.


kutschke (approved):
At the Comp/Soft meeting today we decided to merge this today.  We also created and issue Mu2e/Offline#1157 to follow up on the work suggested in the submit comments of  this  PR.

PR #288, opened 2024-01-25
updated paths of the trigger config files

gianipez:
This PR integrates the use of the mu2e_trig_config repo.


kutschke (approved):
Tested with Offline 1178 and passed tests.

PR #291, opened 2024-02-09
add database to CI tests

rlcee:
Dave is turning on database access in more proditions by default.  This means jobs which use these proditions need to have the database configured.  This PR adds the db configuration to more CI test fcl. (already present some places.) The configuration is in `Production/Validation/database.fcl`
which is included many places in validation fcl.


brownd1978 (approved):

kutschke:
Following a discussion on this thread, Produciton 292 and Offline 1183, are merging both Production 291 and 292 now.  Then retest Offline 1183.

PR #299, opened 2024-02-27
Update mdc2020_primary.cfg

sophiemiddleton:


sophiemiddleton:
this has bothered me for a while, its fixed in my MDC2024 branch and in my local branches but just in case some one else were to use this script

CampaignConfig/mdc2020_primary.cfg
```diff
@@ -25,7 +25,7 @@ nconcat = 1
 pdgcode = -11
 startmom = 0
 endmom=110
-bfield = "Offline/Mu2eG4/geom/bfgeom_DS70_no_tsu_ps_v01.txt"
+bfield = "Offline/Mu2eG4/geom/bfgeom_no_tsu_ps_v01.txt"
```
kutschke, line 28: I am not an expert on these scripts.   Was any production incorrectly run with the DS70 field map? 

Also, when you change the fieldmap, should you also change the particlename parameter?  In my mind those look coupled.

sophiemiddleton: no this was ran a while ago for the special camapign. Its not been used in any recent Production its been overridden in all cases in my local scripts

brownd1978 (approved):

PR #302, opened 2024-03-06
Various Updates from Recent Studies

sophiemiddleton:
This is a large PR incoperating several new things which I think would be best in the main branch:

* Ensemble script updates
* cosmics changes for the recent CORSIKA runs
* updates to digi and reco for the new trigger configuration
* added in the proto fclless scripts, further development will be in MDC2024


kutschke:
@sophiemiddleton Can you recommend an appropriate reviewer for this PR?

CampaignConfig/fclless/mdc2024_primary.cfg
```diff
@@ -0,0 +1,119 @@
+[global]
+includes = /exp/mu2e/app/users/sophie/ProdClone/Production/CampaignConfig/mdc2024_prolog.cfg
+run_number = 1210
```
brownd1978, line 3: A different run number is set in gen_primary.    We only recently identified a case where run numbers are needed in production, to distinguish Extracted from normal geometry sims.  This requires consistency in all the scripts.

CampaignConfig/fclless/run_Primary.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
brownd1978, line 1: This looks like the script running on the worker node, could you add a comment explaining that?

CampaignConfig/fclless/run_Primary.sh
```diff
@@ -0,0 +1,16 @@
...
+echo "args: $@"
+echo "fname=$fname"
+echo "pwd=$PWD"
+echo "ls of default dir"
+ls -al
+IND=$( echo $fname | awk -F. '{print $5}' | sed 's/^0*//' )
+TARF=$(ls $CONDOR_DIR_INPUT/*.tar)
+echo "IND=$IND TARF=$TARF"
+mu2ejobfcl --parfile $TARF --index $IND --default-proto root --default-loc tape > temp.fcl
+echo "$(date) run_Primary temp.fcl content"
```
brownd1978, line 12: Would a different script be needed for (say) run_Mix.sh?  I don't see anything primary-specific about this script.

CampaignConfig/mdc2020_corsika.cfg
```diff
@@ -1,30 +1,31 @@
 [global]
 includes = /exp/mu2e/app/users/sophie/ProductionPush/Production/CampaignConfig/mdc2020_prolog.cfg
-desc = MDC2020aa1
-db_folder = mdc2020aa1
-run_number = 1202
+desc = MDC2020ab
```
brownd1978, line 3: I think we will be at MDC2020ac before this is run...

CampaignConfig/mdc2020_corsika.cfg
```diff
@@ -1,30 +1,31 @@
...
+db_folder = mdc2020ab
+run_number = 1200
```
brownd1978, line 5: Again we need to apply a consistent run numbering, ideally just to separate Extracted and nominal geometry

CampaignConfig/mdc2020_primary.cfg
```diff
@@ -14,21 +14,18 @@ digidb_version = v1_1
 recodb_version = v1_1
 db_purpose = best
 data_tier = dts
-outdir = /pnfs/mu2e/tape/phy-sim/dts/mu2e/
 stopped = Muminus
 evts = 2000
 primary_name = NoPrimary
```
brownd1978, line 19: The name of this file is confusing, given that it's for NoPrimary

CampaignConfig/mdc2020_primary.cfg
```diff
@@ -14,21 +14,18 @@ digidb_version = v1_1
...
 digitype = OnSpill
-beam = 2BB
+beam = 1BB
```
brownd1978, line 21: We eventually need to have 1BB and 2BB NoPrimary.  Maybe they should be separate campaigns?

brownd1978 (commented):
This PR is outside the nightly validation coverage since it brings in new features.  There are a few things you might want to address.  We also need a non-POMS version of fclless production scripts for testing, perhaps some of that content should be moved to Scripts?  Or a new directory?

CampaignConfig/fclless/run_Primary.sh
sophiemiddleton: yes, this is ran on the grid. It should be moved to Scripts eventually

CampaignConfig/fclless/run_Primary.sh
sophiemiddleton: I think probably not given all the parameters are defined in the parfile. So yes I guess this script is more generic than the name suggests but I havent tested it yet.

This is a first step, it will evolve.

CampaignConfig/mdc2020_primary.cfg
sophiemiddleton: no that is always over ridden, I can try to remove this parameter from the global

brownd1978 (approved):
My 2 main issues (run number consistency, directory structure for ensembles scripts) can wait for a follow-on PR.

PR #305, opened 2024-03-15
Update version

brownd1978:


rlcee (approved):
Sorry for not seeing this sooner, I'm often missing the github alerts.  This change will technically work fine, I checked the table list and run coverage.  I assume you are vouching for the content.

PR #307, opened 2024-04-11
Changes in the preparation for S2 corsika run

oksuzian:
This allows to locate files in POMS monitoring stage


CampaignConfig/mdc2020_DIOTail.ini
```diff
@@ -6,7 +6,7 @@ campaign_stage_list = DIOTail_fcl, DIOTail, DIOTaildigi_fcl, DIOTaildigi, DIOTai
 
 [campaign_defaults]
 vo_role=Production
-software_version=Offline
+software_version=mdc2020
```
brownd1978, line 9: If you use MDC2020 you could create the desc field as software_version+release.   That assumes POMS lets you manipulate strings.

JobConfig/cosmic/SpillSplitter.fcl
```diff
@@ -11,32 +11,36 @@ services.SeedService.baseSeed : @local::Common.BaseSeed
 physics: {
   filters : {
     SignalFilter : { # select particles in the Ce signal momentum band, with no cut on the hits
```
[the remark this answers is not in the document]
brownd1978: I think this is OK for now.  We can revisit the names after we've gotten experience using these outputs.

brownd1978 (approved):

Scripts/gen_CosmicStage2.sh
```diff
@@ -112,23 +107,24 @@ if [[ -n $DSSTOPS ]]; then
...
 fi
 
 if [ ! -f $DSSTOPS ]; then
   echo "Can't find DSStops"
   exit_abnormal
 fi
 
-cmd="generate_fcl --dsconf ${OUTCONF} --dsowner ${OWNER} --run-number ${RUNNUM} --description ${S2OUT} --embed ResampleS1.fcl
---events-per-job ${NEVTS} --njobs ${NJOBS} --auxinput=1:physics.filters.CosmicResampler.fileNames:${DSSTOPS}"
+cmd="mu2ejobdef --embed ResampleS1.fcl --setup /cvmfs/mu2e.opensciencegrid.org/Musings/SimJob/${S1CONF}/setup.sh --run-number=${RUNNUM} --events-per-job=${NEVTS} --desc ${S2OUT} --dsconf ${OUTCONF} --auxinput=1:physics.filters.CosmicResampler.fileNames:${DSSTOPS}"
```
brownd1978, line 118: how will you test when developing (ie before there's a musing)?  Will you create a temporary musing?  Nominally things in /cvmfs are considered archival, do we want a temporary location for testing?

brownd1978 (commented):
I think using standard names (MDC2020 instead of mdc2020) would help readability, and might be useful later.

CampaignConfig/mdc2020_DIOTail.ini
oksuzian: I can change:
[sam_consumer]
appvers = %(release)s

However, I don't think I can put the following in *.ini file:
software_version=mdc2020%(release_v_o)s

This will require a user to explicitly set:
software_version=mdc2020ad
That becomes a nuisance to maintain.

Scripts/gen_CosmicStage2.sh
oksuzian: We now test with the latest version of SimJob.
Local scripts have to be uploaded.
We can consider testing against nightly builds.

brownd1978 (approved):

PR #308, opened 2024-04-12
Updated database version to MDC2020_best v1_2

bonventre:
Includes new CRV tables and AlignedTrackerSim tables. AlignedTracker reco tables updated to include sim wire offsets. Should be merged at the same time as corresponding Offline PR enabling AlignedTrackerSim db use

It is expected that reco performance (chi2/dof etc.) will look slightly worse in validation.

Note that trigger validation has also been moved to v1_2 but is running still on v1_0 digis.

### DIFF
--- a/Validation/nightly/extracted_04.fcl
+++ b/Validation/nightly/extracted_04.fcl
@@ -1,6 +1,6 @@
 #include "Production/JobConfig/reco/Extracted.fcl"
 services.DbService.purpose: MDC2020_perfect
-services.DbService.version: v1_0
+services.DbService.version: v1_3
 services.DbService.verbose : 2
 
 #----------------------------------------------------------------
--- a/Validation/nightly/extracted_04.fcl
+++ b/Validation/nightly/extracted_04.fcl
@@ -1,7 +1,5 @@
 #include "Production/JobConfig/reco/Extracted.fcl"
-services.DbService.purpose: MDC2020_perfect
-services.DbService.version: v1_3
-services.DbService.verbose : 2
+#include "Production/Validation/database.fcl"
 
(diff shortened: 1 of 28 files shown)


rlcee (approved):
Thanks for compressing ceMix

Validation/nightly/extracted_04.fcl
```diff
@@ -1,6 +1,6 @@
 #include "Production/JobConfig/reco/Extracted.fcl"
 services.DbService.purpose: MDC2020_perfect
-services.DbService.version: v1_0
+services.DbService.version: v1_3
```
brownd1978, line 3: Why did this jump to v1_3 instead of v1_2?  does v1_3 exist?  Should this come from database.fcl?

bonventre: This is MDC2020_perfect instead of best since it runs on perfect digis right now

brownd1978: Generally we reserve perfect for algorithm debugging.  Is there a reason besides history extracted validation isn't using best?

bonventre: It's only historical, and should be updated soon with a newer list of input files

brownd1978: I just tested that changing extracted_0*.fcl to use MDC2020_best, v1_2 runs without errors and produces reasonable results.  Is there an inconsistency I'm missing?
If we could consolidate all the nightly scripts to use database.fcl now that would be better, even if the plan is to replace them soon.

bonventre: Ok updated for now, will remember that I still need to check on updating these inputs

PR #309, opened 2024-04-18
Use correct run number in CI and use standard script

brownd1978:
This should resolve the problem with CI using a run number outside what the database tables expect.


kutschke (approved):
Ralf's postiive review did not unblock merging.  I will add mine to see if it's just that he's not on the right list?

kutschke:
The repo is configured that only people with write access or better may approve.   We already know that we need to improve our organization of who may approve what.

The CI is still not working following an upgrade of the server supporting Jenkins.   I did a hand test of this merge and verified that the two modified fcl files ran 10 events with a return code of 0.

PR #314, opened 2024-04-30
Cosmic s2

oksuzian:
Final modifications for CORSIKA S2 campaign using MDC2020ae release

### DIFF
--- a/CampaignConfig/mdc2020_corsika.cfg
+++ b/CampaignConfig/mdc2020_corsika.cfg
@@ -35,20 +35,19 @@ submit.expected-lifetime = 28h
 submit.timeout = 27h
 
 [stage_generation]
-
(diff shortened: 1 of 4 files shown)


brownd1978 (approved):

CampaignConfig/mdc2020_corsika.cfg
```diff
@@ -1,18 +1,19 @@
...
+release_v_o = ae
 njobs_s1 = 17500
-njobs = 10
-events_per_job = 5000
+njobs = 82330
+events_per_job = 500000
 s1_start_job=0
-s1type = Low # Only accepts 'Low' or 'All'
+#s1type only accepts 'Low' or 'All'
+s1type = Low
```
brownd1978, line 14: Is s1type still used?  It looks instead as if both Low and All output streams are always defined.

PR #316, opened 2024-05-05
Use specific trigger lines to define Signal

brownd1978:
This insulates simulation production from new triggers which are being added experimentally.


kutschke (approved):
I understand the intent and agree this will work with all of the trigger tables that I know about.

PR #317, opened 2024-05-06
Fix typo in PR 316.

kutschke:
valCompare of 2000 ceSimReco events has an exact match to before the bug was introduced.


brownd1978 (approved):

kutschke:
I look at this as a hotfix and will merge it now so that it is in place for tonight's validation.

PR #318, opened 2024-05-14
Mock data 2024

sophiemiddleton:
a few fixes and updates needed to get mixing of the ensembles to work in latest Production


brownd1978 (approved):

JobConfig/ensemble/python/normalizations.py
```diff
@@ -36,6 +34,31 @@
         target_stopped_mu_per_POT = rate * 1000 
 print(f"Final stops rate {target_stopped_mu_per_POT}")
 
+# get number of POTs in given livetime
+def livetime_to_pot(livetime, run_mode = '1BB'): #livetime in seconds
```
brownd1978, line 38: Its striking how many numbers are needed to produce this result.  These names are clear so it's a good reference.  Beyond the scope of this PR, we should consider how these numbers can be obtrained from a single, authoritative source.

brownd1978 (approved):

PR #319, opened 2024-05-22
Addition of Scripts/projPy_digi.py to submit multiple campaign at once

oksuzian:
An example:
./Scripts/projPy_digi.py --ini_file CampaignConfig/mdc2020_digireco_projectpy.ini --cfg_file CampaignConfig/mdc2020_digireco.cfg --comb_json Scripts/combo_dict.json --create_campaign --submit


Scripts/combo_dict.json
```diff
@@ -0,0 +1,5 @@
+{
```
brownd1978, line 1: Are we going to be creating a lot of these json files?  I suspect so, in which case we should have a dedicated directory for them.  mu2e_trig_config puts json files in the data subdirectory, I suggest that.

brownd1978 (commented):
Unless it's somehow disruptive, I suggest to move the json files to their own directory.

brownd1978:
@FNALbuild run build test

brownd1978 (approved):
a description of the intended use of .cfg, .ini files here or (better) on the wiki would be a helpful addition

PR #325, opened 2024-06-17
Minor fcl file fixes

PawelPlesniak:
The previous directory for the STM pileup scripts was `Production/JobConfig/pileup`, it's now `Production/JobConfig/pileup/STM`

### DIFF
--- a/Scripts/gen_BeamToVD101.sh
+++ b/Scripts/gen_BeamToVD101.sh
@@ -70,8 +70,8 @@ for dirname in 000 001 002 003 004 005 006 007 008 009; do
 done
 
 # Cleanup
-#echo "Removing produced files"
-#rm -f beamToVD101Resampler.fcl
-#rm -f EleBeamCat.txt
-#rm -f MuBeamCat.txt
-#echo "Finished"
+echo "Removing produced files"
+rm -f beamToVD101Resampler.fcl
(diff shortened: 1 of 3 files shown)


Scripts/gen_BeamToVD101.sh
```diff
@@ -70,8 +70,8 @@ for dirname in 000 001 002 003 004 005 006 007 008 009; do
 done
 
 # Cleanup
-echo "Removing produced files"
-rm -f beamToVD101Resampler.fcl
-rm -f EleBeamCat.txt
-rm -f MuBeamCat.txt
-echo "Finished"
+#echo "Removing produced files"
```
kutschke, line 73: Just checking: did you mean to leave this commented out?   Unless commented out code is debug code or code under development we would like it to be removed.

kutschke (approved):
Looks good .  One stylist comment.

Scripts/gen_BeamToVD101.sh
PawelPlesniak: Thanks for catching that Rob. That was used for debugging which is now resolved. I have pushed the change.

kutschke (approved):
Thanks for looking after the comment out code.

kutschke:
I don't believe that these changes have CI coverage.  So I am not rerunning the CI.

PR #327, opened 2024-06-24
update IPAMuminusMichel primary generator and prefilter

edcallaghan:
This PR updates the configuration for `IPAMuminusMichel` primary jobs to use a new (and as-yet unmerged) generator which accommodates heterogeneous materials. The configuration implemented here samples muons which stopped in the IPA volume, and chooses which element the muon captured into according to the relative atomic volumes of carbon and hydrogen. The chosen element then defines the decay-time distribution and DIO spectrum associated with the most-abundant nucleus, respectively. The fcl files produced using `gen_Primary.sh` have been tested to run without error, but depend on https://github.com/Mu2e/Offline/pull/1290.

This PR also makes use of Sophie's `GenFilter` to prefilter out electrons which have trajectory which will never intersect a tracker straw, which will lead to savings in cpu consumption.

### DIFF
--- a/JobConfig/primary/IPAMuminusMichel.fcl
+++ b/JobConfig/primary/IPAMuminusMichel.fcl
@@ -5,14 +5,13 @@
 # original author: S Middleton
 #
 #include "Production/JobConfig/primary/IPAStopParticle.fcl"
-
 physics.filters.PrimaryFilter.MinimumPartMom : 1.0
 physics.producers.FindMCPrimary.PrimaryProcess : "mu2eMuonDecayAtRest"
 
-CompositeGenerator: {
+physics.producers.generate: {
   module_type: CompositeMaterialGenerator
-  tag: IPAMuminusStopResampler
-  process: "mu2eMuonDecayAtRest"
+  SimParticleCollection: IPAMuminusStopResampler
+  processCode: "mu2eMuonDecayAtRest"
   weighting: {
     tool_type: AtomicVolumeSamplerTool
     material: "IPAPolyethylene"
@@ -56,7 +55,5 @@ CompositeGenerator: {
     }
   ]
 }
-
-physics.producers.generate: @local::CompositeGenerator
 
 outputs.PrimaryOutput.fileName: "dts.owner.IPAMuminusMichel.version.sequencer.art"
--- a/JobConfig/primary/IPAMuminusMichel.fcl
+++ b/JobConfig/primary/IPAMuminusMichel.fcl
@@ -13,7 +13,7 @@ physics.producers.generate: {
   SimParticleCollection: IPAMuminusStopResampler
   processCode: "mu2eMuonDecayAtRest"
   weighting: {
-    tool_type: AtomicVolumeSamplerTool
+    tool_type: BoundMuonDecayFractionSamplerTool
     material: "IPAPolyethylene"
   }
   elements: [
--- a/JobConfig/primary/IPAMuminusMichel.fcl
+++ b/JobConfig/primary/IPAMuminusMichel.fcl
@@ -6,16 +6,57 @@
 #
 #include "Production/JobConfig/primary/IPAStopParticle.fcl"
 
-physics.filters.PrimaryFilter.MinimumPartMom : 30.0 // MeV/c TODO - tweak this parameter to get more DetSteps
+physics.filters.PrimaryFilter.MinimumPartMom : 1.0
 physics.producers.FindMCPrimary.PrimaryProcess : "mu2eMuonDecayAtRest"
 
-physics.producers.generate : {
-  module_type : Pileup
-  verbosity : 0
-  inputSimParticles: IPAMuminusStopResampler
-  captureProducts : []
-  decayProducts : [ @local::Pileup.dioGenTool ]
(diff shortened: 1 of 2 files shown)


JobConfig/primary/IPAMuminusMichel.fcl
```diff
@@ -6,16 +6,57 @@
...
 physics.producers.FindMCPrimary.PrimaryProcess : "mu2eMuonDecayAtRest"
 
-physics.producers.generate : {
-  module_type : Pileup
-  verbosity : 0
-  inputSimParticles: IPAMuminusStopResampler
-  captureProducts : []
-  decayProducts : [ @local::Pileup.dioGenTool ]
-  stoppingTargetMaterial : "IPA"
+CompositeGenerator: {
```
kutschke, line 12: The base definition of a module configuration belongs in Offline.  In Production, reference it with "@table::" here and override anything that needs to be overridden.  Just as you do for GenFilter, below.

JobConfig/primary/IPAMuminusMichel.fcl
```diff
@@ -6,16 +6,57 @@
...
+  module_type: CompositeMaterialGenerator
+  tag: IPAMuminusStopResampler
```
kutschke, line 14: When you move this to Offline/EventGenerator/fcl/prolog.fcl, set this to have a value of @nil.  If someone forgets to override this it will produce a run-time error when someone tries to read the parameter - so it fails loudly and safely.  

My own style is to never label something just "tag" - give it a descriptive name.

JobConfig/primary/IPAMuminusMichel.fcl
```diff
@@ -6,16 +6,57 @@
...
+  process: "mu2eMuonDecayAtRest"
```
kutschke, line 15: I would also decorate "process" with an adjective.  It's a processCode .

JobConfig/primary/IPAMuminusMichel.fcl
```diff
@@ -6,16 +6,57 @@
...
+    {
+      name: "H"
+      position_tool: {
+        tool_type: MuStopDecayPositionSamplerTool
+        atom: "H"
+      }
+      generator_tool: {
+        tool_type: DIOGenerator
+        spectrum: {
+          pdgId: 11
```
kutschke, line 48: Does your code read this or someone else's?  

If it's your code please use the name of the particle, not it's number.   You can construct an object of type PDGCode from a string and it will scan the known strings to find the number.

kutschke (changes requested):

JobConfig/primary/IPAMuminusMichel.fcl
edcallaghan: All of the analogous primary generators are configured directly as definition of `physics.producers.generate`. Here I defined it as a standalone object (? I'm not sure the right terminology here) and then defined `physics.producers.generate` as that object, which I guess does violate the rule you cite. I've now switched to directly define `physics.producers.generate`, consistent with the other primaries.

JobConfig/primary/IPAMuminusMichel.fcl
edcallaghan: The latest push in https://github.com/Mu2e/Offline/pull/1290 exchanges `tag` for `SimParticleCollection`.

JobConfig/primary/IPAMuminusMichel.fcl
edcallaghan: The latest push in https://github.com/Mu2e/Offline/pull/1290 exchanges `process` for `processCode`.

JobConfig/primary/IPAMuminusMichel.fcl
edcallaghan: The blocks containing the `pdgId` fields are not read by the new generator --- they are only seen as `fhicl::DelegatedParameter`s at the generator level. The code which reads the `pdgId` field here is [the preexisting DIOGenerator tool](https://github.com/Mu2e/Offline/blob/main/EventGenerator/src/DIOGenerator_tool.cc).

kutschke: Thanks.  This is distant enough from the purpose of this PR that we can defer the change to a later PR

JobConfig/primary/IPAMuminusMichel.fcl
kutschke: Thanks - I see the matching change here.

JobConfig/primary/IPAMuminusMichel.fcl
kutschke: Thanks - I see the matching change here.

JobConfig/primary/IPAMuminusMichel.fcl
kutschke: Ok.  I see that they all do it that way.  Another thing to fix in a later PR.  The reason we would like the base fcl definition defined in ../fcl/prolog.fcl relative to the module source code is to minimize multiple points of maintenance.

kutschke:
@FNALbuild run build test with Mu2e/Offline#1290

kutschke (approved):
Thanks for making in updates.

PR #328, opened 2024-07-08
Latest progress on ensembles scripts

sophiemiddleton:
PR contains new scripts to help produce ensembles


brownd1978 (approved):

sophiemiddleton:
I fixed those merge conflicts, sorry for delay

brownd1978 (commented):
If you have a dedicated gen_primary2024 I don't see why you need to modify gen_Primary, since the only use case I can see for that mod is generating MDC2024. Am I missing something?

sophiemiddleton:
Dave, there is an overlap....but the difference is that the gen_Primary.sh still uses gen fcl while gen_Primary2024 was part of my original work doing the fclless, so it uses jobdef. Of course gen_Primary should also be updated in the same way. gen_Primary2024....that is already in Mu2e  but inside the fclless dir.  Iwill remove it, in my next round I will change gen_Primary to use the fclless tools

Scripts/gen_Primary.sh
```diff
@@ -3,13 +3,14 @@
...
-# Usage: ./Production/Scripts/generate_Primary.sh --primary CeEndpoint --campaign MDC2020 --pver v --sver p --type Muminus --njobs 1000 --events 4000 --pdg 11 --start 0 --end 110 --field Offline/Mu2eG4/geom/bfgeom_reco_altDS11_helical_v01.txt
+# Usage: ./Production/Scripts/generate_Primary.sh --primary CeEndpoint --PCAMPAIGN MDC2020 --pver v --sver p --type Muminus --njobs 1000 --events 4000 --pdg 11 --start 0 --end 110 --field Offline/Mu2eG4/geom/bfgeom_reco_altDS11_helical_v01.txt
 #
 # Note: User can omit flat (pdg, startmom and enedmom) arguments without issue. Field argument also generally will not be used
 
-# The main input parameters needed for any campaign
+# The main input parameters needed for any PCAMPAIGN
 PRIMARY="" # is the primary
-CAMPAIGN="" # Campaign (MDC2020"
+PCAMPAIGN="" # primary CAMPAIGN (MDC2020"
```
brownd1978, line 12: please revert the name change (pcampaign->campaign) and make scampaign optional with default value=campaign. That maintains backwards-compatibility and simplifies the interface.

JobConfig/ensemble/fcl/SamplingInput_sr0.fcl
```diff
@@ -0,0 +1,53 @@
...
+source : { module_type : SamplingInput
+  inputCommands : @local::Ensemble.InputCommands
+  dropDescendantsOfDroppedBranches: false
+  run : 1201
+  subRun : 0
+  samplingSeed : 1
+  dataSets : { 
+      CeMLL: {
+        fileNames : ["/pnfs/mu2e/tape/phy-sim/dts/mu2e/CeMLeadingLog/MDC2024a_sm4/art/01/1d/dts.mu2e.CeMLeadingLog.MDC2024a_sm4.001210_00000691.art"]
+        weight : 1.088855e-06
```
brownd1978, line 22: outside the scope of this PR we need to make these numbers come from the database, ala simefficiencies. They are campaign-dependent and we don't want to hardcode them in any script. Please create an issue for that after this PR is merged.

JobConfig/ensemble/fcl/SamplingInput_sr0.fcl
```diff
@@ -0,0 +1,53 @@
...
+
+source : { module_type : SamplingInput
+  inputCommands : @local::Ensemble.InputCommands
+  dropDescendantsOfDroppedBranches: false
+  run : 1201
+  subRun : 0
+  samplingSeed : 1
+  dataSets : { 
+      CeMLL: {
+        fileNames : ["/pnfs/mu2e/tape/phy-sim/dts/mu2e/CeMLeadingLog/MDC2024a_sm4/art/01/1d/dts.mu2e.CeMLeadingLog.MDC2024a_sm4.001210_00000691.art"]
```
brownd1978, line 21: If this filename is intended as an example please put it in comments. We don't want to hard-code any file names in production scripts.

JobConfig/ensemble/python/normalizations.py
```diff
@@ -131,7 +146,7 @@ def cry_offspill_normalization(livetime, run_mode = '1BB'):
     if(run_mode == '2BB'):
       # 2BB
       offspill_dutyfactor = 0.246
```
brownd1978, line 148: here too we need a central location for common numbers. Analyzers will need to know these as well. It should become an issue.

brownd1978 (changes requested):
Some minor changes requested. This PR breaks a lot of new ground and is fine as a development point but it also raises long-term questions that need to be addressed in a more structural way. Those should be made into issues. I'm also requesting rbonventre to review the python scripts.

JobConfig/ensemble/fcl/SamplingInput_sr0.fcl
sophiemiddleton: this is just and example. one of my new scripts will make this fcl

sophiemiddleton: (I mean it will make the updated version)

JobConfig/ensemble/fcl/SamplingInput_sr0.fcl
sophiemiddleton: its just an example

JobConfig/ensemble/python/normalizations.py
sophiemiddleton: yes agreed

sophiemiddleton:
I will remove the example fcl since it seems to be causing issues.

sophiemiddleton:
Hi Dave, I will be away at ICHEP until 26th July following the review. I will get back to this PR when I return.

JobConfig/ensemble/python/normalizations.py
sophiemiddleton: https://github.com/Mu2e/Production/issues/334

Scripts/gen_Primary.sh
sophiemiddleton: done

sophiemiddleton:
@brownd1978 @bonventre I have carried out Dave's requested changes...Richie did you have any further suggestions. I will draft out a table design and send it to you.

I will soon put in another PR with changes for the RPC campaign

brownd1978:
@FNALbuild run build test

sophiemiddleton:
Any updates on merging this?

brownd1978 (approved):
thanks for making the updates

brownd1978:
@FNALbuild run build test

PR #331, opened 2024-07-10
surface step compression

brownd1978:
Patch on top of PR #330 to address CI issues

### DIFF
--- a/JobConfig/pileup/prolog.fcl
+++ b/JobConfig/pileup/prolog.fcl
@@ -22,6 +22,7 @@ Pileup: {
       module_type : CompressDetStepMCs
       strawGasStepTag : "StrawGasStepMaker"
       caloShowerStepTag : "CaloShowerStepMaker"
+      surfaceStepTag : ""
       crvStepTag : "CrvSteps"
       simParticleTags : [ "g4run" ]
       debugLevel : 0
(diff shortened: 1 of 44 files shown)


JobConfig/pileup/prolog.fcl
```diff
@@ -22,6 +22,7 @@ Pileup: {
       module_type : CompressDetStepMCs
       strawGasStepTag : "StrawGasStepMaker"
       caloShowerStepTag : "CaloShowerStepMaker"
+      surfaceStepTag : ""
```
AndrewEdmonds11, line 25: Thanks, Dave. I think this should be ```surfaceStepTags : [ ]```

AndrewEdmonds11 (commented):
Thanks, Dave. I think just one correction noted below. Also, should I close #330 and then run Mu2e/Offline#1306 tests with this one?

JobConfig/pileup/prolog.fcl
brownd1978: Hmm, it worked fine when I tested it. The code seems to require an atom, not a sequence. Maybe you're thinking of CompressDigiMCS?

brownd1978:
@FNALbuild run build tests with Mu2e/Offline#1306

AndrewEdmonds11 (approved):

PR #333, opened 2024-07-26
fix missing values for validation

brownd1978:


kutschke (approved):
Thanks for looking after this.

PR #337, opened 2024-08-06
fix normalization

sophiemiddleton:
- bug discovered in normalization script, this is a fix


brownd1978 (approved):

sophiemiddleton:
I think we have verified these numbers now.

PR #338, opened 2024-08-06
Fix some bugs

brownd1978:
Correct the example invocation. Replace samweb with with samListLocations, as the former gives all locations (tape and persistent, tape first) while the latter preferentialy picks the persistent location if available.

This script needs more work: replacing generate_fcl with mu2ejobdef, etc. I also don't see the need for the 'tag' variable, and I suggest getting rid of the 2nd campaign instance. I think keeping the campaign the same as the one used to generate stopped particles will make it easier for consumers to understand the provenance. These should be addressed in subsequent PRs.

### DIFF
--- a/Scripts/gen_Primary.sh
+++ b/Scripts/gen_Primary.sh
@@ -48,7 +48,7 @@ usage() {
   [ --run (opt) default 1202 ]
   [ --cat(opt) default Cat ]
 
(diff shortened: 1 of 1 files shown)


kutschke:
@brownd1978 Is this an example of a PR that was automatically marked draft but should be marked ready for review?  I read all of your second para as work for future PRs.  Is that what you meant?

sophiemiddleton (approved):
looks good

PR #341, opened 2024-08-19
Fixing Primary to work with jitfcl

oksuzian:


Scripts/gen_Primary.sh
```diff
@@ -195,17 +178,26 @@ if [[ "${FLAT}" == "FlatMuDaughter" ]]; then
...
+parfile=$(ls cnf.*.tar)
+# Remove cnf.
+index_dataset=${parfile:4}
+# Remove .0.tar
+index_dataset=${index_dataset::-6}
+
+idx_format=$(printf "%07d" ${JOBS})
+echo $idx
+echo "Creating index definiton with size: $idx"
+samweb create-definition idx_${index_dataset} "dh.dataset etc.mu2e.index.000.txt and dh.sequencer < ${idx_format}"
```
brownd1978, line 201: I suggest registering in sam to be optional, otherwise it will occur even when running tests. Maybe add a production mode where this is on by default?

Scripts/gen_Primary.sh
```diff
@@ -195,17 +178,26 @@ if [[ "${FLAT}" == "FlatMuDaughter" ]]; then
...
-  --embed primary.fcl --auxinput=1:physics.filters.${resampler}.fileNames:Stops.txt
-for dirname in 000 001 002 003 004 005 006 007 008 009; do
-  if test -d $dirname; then
-    echo "found dir $dirname"
-    rm -rf ${PRIMARY}\_$dirname
-    mv $dirname ${PRIMARY}\_$dirname
-    echo "moving $dirname to ${PRIMARY}_${dirname}"
-  fi
-done
+cmd="mu2ejobdef --embed primary.fcl --setup ${SETUP} --run-number=${RUN} --events-per-job=${EVENTS} --desc ${PRIMARY} --dsconf ${PRIMARY_CAMPAIGN} --auxinput=1:physics.filters.${resampler}.fileNames:Stops.txt"
```
brownd1978, line 187: mu2ejobdef assumes you are running in a unique directory: the output tar file is always produced in the workdir, and it will combine the results from any previous invocation of mu2ejobdef. I suggest to delete any previous .tar files in this directory to avoid confusion, at least in production mode. Ideally mu2ejobdef would take care of this itself: I have suggested it to the author but did not receive a positive response.

brownd1978: Outside the scope of this PR: Do we want to consider archiving the tar file in 'production' mode? Maybe making a copy on persistent?

brownd1978 (commented):
I'd like to understand the interaction with sam better. otherwise this all looks good.

Scripts/gen_Primary.sh
```diff
@@ -26,16 +25,14 @@ ENDMOM=110 # optional (for flat only)
...
 RUN=1202
 CAT="Cat"
-TAG=""
 
 # Function: Print a help message.
 usage() {
   echo "Usage: $0
   [ --primary primary physics name ]
-  [ --campaign primary camapign name ]
-  [ --scampaign stops camapign name ]
```
sophiemiddleton, line 36: Hi Yuri, we need to retain the possibility of different campaigns for in and output

Scripts/gen_Primary.sh
```diff
@@ -146,45 +140,34 @@ else
...
 echo physics.filters.${resampler}.mu2e.MaxEventsToSkip: ${nskip} >> primary.fcl
 echo "services.GeometryService.bFieldFile : \"${FIELD}\"" >> primary.fcl
+echo outputs.PrimaryOutput.fileName: \"dts.owner.${PRIMARY}.version.sequencer.art\"  >> primary.fcl
 
-if [[ "${TAG}" == "DIOtail" ]]; then
-  echo "DIOTail"
+if [[ "${PRIMARY}" == "DIOtail" ]]; then
   echo physics.producers.generate.decayProducts.spectrum.ehi: ${ENDMOM}        >> primary.fcl
   echo physics.producers.generate.decayProducts.spectrum.elow: ${STARTMOM}    >> primary.fcl
-  echo outputs.PrimaryOutput.fileName: \"dts.owner.${PRIMARY}.version.sequencer.art\"  >> primary.fcl
```
sophiemiddleton, line 187: I think I understand why this was encapsulated. Since in the DIOtail "primary name" becomes DIOtailp95MeVc etc. and we need to keep that part of the name in the dts file. So it needs to be within there. We need to retain the "tag" as DIOtail so we can reuse the .fcl input for both

sophiemiddleton: It doesnt matter which you choose but the point is the .fcl is multi purpose is can be used for any DIO startmom but the file name must have the start mom in there

CampaignConfig/mdc2020_primary.ini
```diff
@@ -0,0 +1,65 @@
...
+poms_role = production
+name = mdc2020_primary_081924
+campaign_stage_list = generation_par, generation
+
+[campaign_defaults]
+vo_role=Production
+campaign_keywords={"release_v_dts": "ag", 
+		   "release_v_stops": "p",
+		   "stopped_type": "Muminus",
+		   "primary_name": "DIOtail",
```
sophiemiddleton, line 12: We need to follow my comment or design a new way to include the mom start value in the name. The "Tag" parameter was used to do this previously

Scripts/gen_Primary.sh
oksuzian: >I suggest to delete any previous .tar files in this directory to avoid confusion
Will do

CampaignConfig/mdc2020_primary.ini
```diff
@@ -0,0 +1,66 @@
...
+campaign_stage_list = generation_par, generation
+
+[campaign_defaults]
+vo_role=Production
+campaign_keywords={"release_v_dts": "ag", 
+		   "release_v_stops": "p",
+		   "stopped_type": "Muminus",
+#		   "primary_name": "DIOtail_95",
+		   "primary_name": "CeEndpoint",
+		   "start_mom": "95",
```
sophiemiddleton, line 14: thanks for this, I assume the start and end mom are only ever used for DIOtail?

sophiemiddleton (commented):
thanks Yuri, I like the solution. As mentioned I would like to follow your new workflow to learn the process, but I think this looks good

Scripts/gen_Primary.sh
```diff
@@ -109,34 +106,36 @@ while getopts ":-:" options; do
         run)
           RUN=${!OPTIND} OPTIND=$(( $OPTIND + 1 ))
           ;;
+        setup)
+          SETUP=${!OPTIND} OPTIND=$(( $OPTIND + 1 ))
+          ;;
+#        prod)
```
brownd1978, line 112: Is prod commented out because this is still in testing?

Scripts/gen_Primary.sh
```diff
@@ -195,17 +184,28 @@ if [[ "${FLAT}" == "FlatMuDaughter" ]]; then
...
-for dirname in 000 001 002 003 004 005 006 007 008 009; do
-  if test -d $dirname; then
-    echo "found dir $dirname"
-    rm -rf ${PRIMARY}\_$dirname
-    mv $dirname ${PRIMARY}\_$dirname
-    echo "moving $dirname to ${PRIMARY}_${dirname}"
-  fi
-done
+if [[ "$PROD" = true ]]; then
+    rm cnf.*.tar
```
brownd1978, line 194: I thought 'prod' was to trigger sam registration, do you now plan to organize that some other way?
Since prod option is commented out this I don't think this line can actually do anything.
Is there a reason not to rm cnf.*.tar even without 'prod==true'? The following command won't work if there is a pre-existing file.

brownd1978 (commented):
I'm not sure if this is ready or not.

Scripts/gen_Primary.sh
oksuzian: I will remove it.
It's defined and set to true in prolog
Since prolog is used on the grid only that achieves the designed feature.

Scripts/gen_Primary.sh
oksuzian: This will only work on grid.
I don't want to delete user's file by default

sophiemiddleton (approved):

Scripts/gen_Primary.sh
```diff
@@ -25,17 +24,16 @@ STARTMOM=0 # optional (for flat only)
 ENDMOM=110 # optional (for flat only)
 OWNER=mu2e
 RUN=1202
+#PROD=false
```
brownd1978, line 27: more cleanup here...

Scripts/gen_Primary.sh
oksuzian: Removed it

sophiemiddleton:
@brownd1978 can we merge this? I want to get the MDS1 samples made this week

brownd1978 (approved):

PR #347, opened 2024-09-12
Follow parameter name change

brownd1978:
Follow change to make parameter name less misleading. This PR is paired with Offline 1336


brownd1978:
@FNALbuild run build test

bonventre (approved):

PR #350, opened 2024-09-26
now generating deep copies of the collection with the objects triggering the event

gianipez:
tagging: @brownd1978 @rlcee

### DIFF
--- a/Validation/trigger.fcl
+++ b/Validation/trigger.fcl
@@ -42,3 +42,33 @@ physics.trigger_paths : [@sequence::Trig_valMenu.trigger_paths]
 physics.producers.EWMProducer.SpillType : 1
 #include "Production/JobConfig/common/epilog.fcl"
 #include "Production/JobConfig/reco/drop_trigger.fcl"
+
+
+physics.producers.aprHelixTriggerInfoMerger.ntupMode                : 1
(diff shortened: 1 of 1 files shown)


gianipez:
Once this PR will be in, we will have to re-generate a reference file as there will be additional collections producing histograms

Validation/trigger.fcl
```diff
@@ -42,3 +42,33 @@ physics.trigger_paths : [@sequence::Trig_valMenu.trigger_paths]
 physics.producers.EWMProducer.SpillType : 1
 #include "Production/JobConfig/common/epilog.fcl"
 #include "Production/JobConfig/reco/drop_trigger.fcl"
+
+
+physics.producers.aprHelixTriggerInfoMerger.ntupMode                : 1
```
brownd1978, line 47: Didnt the name of that parameter change? Is a change in this PR to follow that needed?

brownd1978 (commented):
Are these deep copies related to the increased resource usage seen in the nightly trigger validation?

Validation/trigger.fcl
gianipez: thanks for spotting this. I simply forgot the last `git push`. Now everything should be fine. I tested it locally successfully.

brownd1978 (approved):

PR #353, opened 2024-10-17
added option to deal with older CrvDigis in Validation

ehrlich-uva:
- old digis were produced relative to -protonBunchTime
- new digis are produced relative to -protonBunchTime-eventTiming.timeFromProtonsToDRMarker

The fcl setting `physics.producers.CrvRecoPulses.oldDigis : true` allows us to read old digis with the new code. This setting is used for the nightly validation job "reco", so that we can continue to use the CrvDigis of the MDC2020 files.


ehrlich-uva:
I'm going to reintroduce the 200ns delay (eventTiming.timeFromProtonsToDRMarker). So this PR is not needed anymore.

PR #360, opened 2024-11-09
explicit dts InputTags for tracker digi maker configuration

edcallaghan:
This PR adapts the configuration of the tracker digi maker, i.e. "`makeSD`," to that implemented in https://github.com/Mu2e/Offline/pull/1366. This is a preparatory change for the implementation of runtime-filtering of mixed detector steps.

### DIFF
--- a/JobConfig/mixing/epilog.fcl
+++ b/JobConfig/mixing/epilog.fcl
@@ -7,7 +7,7 @@ physics.filters.NeutralsFlashMixer.mu2e.simStageEfficiencyTags: ["NeutralsCat", 
 physics.filters.MuStopPileupMixer.mu2e.simStageEfficiencyTags: ["MuBeamCat", "MuminusStopsCat", "MuStopPileupCat"]
 physics.filters.MuStopPileupMixer.mu2e.meanEventsPerPOTFactors : [ @local::Pileup.MuminusTargetStopPrescale ]
 # adjust detstep lists to include pileup
-physics.producers.makeSD.StrawGasStepModule : ""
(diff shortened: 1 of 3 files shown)


JobConfig/mixing/epilog.fcl
```diff
@@ -7,7 +7,7 @@ physics.filters.NeutralsFlashMixer.mu2e.simStageEfficiencyTags: ["NeutralsCat",
 physics.filters.MuStopPileupMixer.mu2e.simStageEfficiencyTags: ["MuBeamCat", "MuminusStopsCat", "MuStopPileupCat"]
 physics.filters.MuStopPileupMixer.mu2e.meanEventsPerPOTFactors : [ @local::Pileup.MuminusTargetStopPrescale ]
 # adjust detstep lists to include pileup
-physics.producers.makeSD.StrawGasStepModule : ""
+physics.producers.makeSD.StrawGasStepModules : [ "EleBeamFlashMixer", "MuBeamFlashMixer", "NeutralsFlashMixer", "MuStopPileupMixer" ]
```
brownd1978, line 10: This list is the same as @sequence::Mixing.StepMixSequence, please use that instead.
I think you still need compressDetStepMCs in the list, otherwise you will miss the DetectorSteps coming from the primary particle (main event stream).

brownd1978 (changes requested):
Overall this is a good improvement/standardization, I think it needs a minor fix.

brownd1978 (approved):
looks good

JobConfig/mixing/epilog.fcl
edcallaghan: Both components are addressed in the latest push, and validation in `Offline` was successful.

PR #367, opened 2024-11-14
Add RMC generation fcl and test fcl, add photon conversion resampling

michaelmackenzie:
Adding generation and detector steps fcl for relevant RMC datasets and photon conversion resampling. This is in connection with [Offline PR 1371](https://github.com/Mu2e/Offline/pull/1371). The RMC generation can be tested with fcl added to `Tests/`, where for photon resampling a conversion point dataset must first be generated using `Tests/RMCGammaWriter.fcl`.

I'm sure I didn't follow the proper style in many places, so please let me know where the fcl should be better organized or if definitions should be moved to alternate places!


JobConfig/primary/IPAStopParticle.fcl
```diff
@@ -8,6 +8,7 @@
 physics.filters.TrackerRadialGenFilter: {
   @table::GenFilter
   isNull: false
+  filterMaxR : true
```
brownd1978, line 11: For future reference, to make a cut optional you can use art::OptionalAtom<> to encode both the value and whether the cut should be applied. That avoids having a separate bool and the possibility of a self-inconsistent config (value set but boolean not).

JobConfig/primary/RMC.fcl
```diff
@@ -0,0 +1,25 @@
+#
+# Radiative muon capture (RMC) spectrum using the Closure Approximation as measured by TRIUMF (1999)
```
brownd1978, line 2: Do we need additional generator codes to distinguish RMC from the closure approximation vs tail generation to the kinematic limit?  ie between known and unknown physics?

JobConfig/primary/RMCExternal.fcl
```diff
@@ -0,0 +1,9 @@
+#
+# RMC External, kmax = 90.1 MeV
+#
+
+#include "Production/JobConfig/primary/RMC.fcl"
+
+physics.producers.generate.decayProducts.spectrum.external : true
+physics.producers.FindMCPrimary.PrimaryProcess : "mu2eExternalRMC"
+outputs.PrimaryOutput.fileName : "dts.owner.RMCExternal.version.sequencer.art"
```
brownd1978, line 9: empty line

JobConfig/primary/RMCGammaEndpointWriter.fcl
```diff
@@ -0,0 +1,11 @@
+#
+# RMC External photons, kmax = 101.9 MeV, saving events with a photon conversion
+#
+
+#include "Production/JobConfig/primary/RMCGammaWriter.fcl"
+
+physics.producers.generate.decayProducts.spectrum.ehi      : 101.9
```
brownd1978, line 7: These numbers also appear in RMCExternalEndpoint, can those scripts be nested to avoid the duplication?

JobConfig/primary/RMCInternalEndpoint.fcl
```diff
@@ -0,0 +1,8 @@
+#
+# RMC Internal, kmax = 101.9 MeV
+#
+
+#include "Production/JobConfig/primary/RMCInternal.fcl"
+physics.producers.generate.decayProducts.spectrum.ehi      : 101.9
```
brownd1978, line 6: more cases of repeated numbers and config.
If nesting doesn’t work could the value at least be set in prolog?

Tests/FlatGammaCaloSteps.fcl
```diff
@@ -0,0 +1,5 @@
+#include "Production/JobConfig/primary/FlatGammaCalo.fcl"
+source.firstRun: 501
```
brownd1978, line 2: This run number isn’t inside the valid MDC2020 range and the results will not be processable (digitization and reconstruction) due to lack of conditions data. I’m not sure if that matters for this test.

Tests/RMCInternalSteps.fcl
```diff
@@ -0,0 +1,12 @@
+#include "Production/JobConfig/primary/RMCInternal.fcl"
+source.firstRun: 501 #TODO
+physics.filters.TargetStopResampler.fileNames : [
+"/pnfs/mu2e/persistent/datasets/phy-sim/sim/mu2e/MuminusStopsCat/MDC2020p/art/6d/cf/sim.mu2e.MuminusStopsCat.MDC2020p.001201_00000000.art"]
```
brownd1978, line 4: I appreciate having test scripts, but there is a lot of replication. When the stops file evolves (which it will) this will require a lot of maintenance to update. Can the repeated parts be consolidated?

brownd1978 (commented):
Overall this looks good, I point out a few targets for improvement.

JobConfig/primary/IPAStopParticle.fcl
michaelmackenzie: Interesting, I didn't realize this! I can make a quick PR to fix this, otherwise I'll make sure to do this next time.

JobConfig/primary/RMC.fcl
michaelmackenzie: This is a fair question. I'm not sure what's the best choice here, it seems reasonable to have the tail measurement connected to its own process code so I'm happy to implement one if recommended.

JobConfig/primary/RMCExternal.fcl
michaelmackenzie: I added a line here.

JobConfig/primary/RMCGammaEndpointWriter.fcl
michaelmackenzie: I moved the measured and kinematic limit endpoints to the prolog and then reference them here, so there's now only one maintenance point if we want to update the values. Does this seem like a better setup?

JobConfig/primary/RMCInternalEndpoint.fcl
michaelmackenzie: I moved this to the prolog.

Tests/FlatGammaCaloSteps.fcl
michaelmackenzie: Is there a better run number to use for testing? I took this from an existing test script.

Tests/RMCInternalSteps.fcl
michaelmackenzie: I moved the muon stop file info (and the run number) to a test fcl file `Production/Tests/MuonStopConfig.fcl` so the muon stop file can more easily be updated. I also updated the CE, DIO, and flat e+/e- test fcl to include this config file to reduce maintenance.

Tests/FlatGammaCaloSteps.fcl
sophiemiddleton: you need something 1200-1210

JobConfig/primary/IPAStopParticle.fcl
brownd1978: If you can patch this quickly in Offline that would be best.

JobConfig/primary/RMC.fcl
brownd1978: Lets defer this to when you have a 'RMC tail' generator.

Tests/FlatGammaCaloSteps.fcl
brownd1978: But don't use 1205, that's for the extracted position.

brownd1978 (approved):
Thanks for the updates. I'm approving as complete, if you want to update to use OptionalAtom now that's fine. I'd like at least 1 other person (Sophie, Yuri) to review before merging.

Tests/FlatGammaCaloSteps.fcl
michaelmackenzie: I updated the test fcl to use run 1210 instead

PR #368, opened 2024-11-20
Add RMC to the muon stop pileup capture products, and add a test script

michaelmackenzie:
Add RMC  the list of muon stop pileup capture products. The RMC rate for aluminum is 1.4e-5 for E > 57 MeV over the closure approximation spectrum fraction for E > 57 MeV (~1/10,000 muon captures). This is in connection to [Offline PR 1375](https://github.com/Mu2e/Offline/pull/1375).


JobConfig/pileup/MuStopPileup.fcl
```diff
@@ -54,6 +54,19 @@ physics : {
             spectrumShape: "flat"
           }
           tool_type: "MuCapPhotonGenerator"
+        },
```
brownd1978, line 57: This configuration is for 'external', does that mean it generates photons? Is that redundant with the flat spectrum above?
What about internal conversion? Should there be a 2nd tool instance for that?

brownd1978 (commented):
Thanks for the quick response! I have some basic questions. Also note that this change won't affect existing MDC2020 output, we would need to remake the pileup stream for it to have an effect. It would be useful to have a comparison of the output before/after this change.

JobConfig/pileup/MuStopPileup.fcl
michaelmackenzie: Yes, we would need to have a second tool in the list for internal conversions. I can add this as well if this makes sense to you. Its rate will be reduced by 0.0069 with respect to the real photons for reference. I can also consider updating the RMC generator tool to randomly generate internal instead of external conversions at this relative rate if this makes more sense.

michaelmackenzie:
> >  Yes, we would need to have a second tool in the list for internal conversions. I can add this as well if this makes sense to you. Its rate will be reduced by 0.0069 with respect to the real photons for reference. I can also consider updating the RMC generator tool to randomly generate internal instead of external conversions at this relative rate if this makes more sense.

> I think we should be generating both internally converted (e+e- pairs) and real photons. Only having 1 seems physically inconsistent. Whether that's 1 tool or 2 is an implementation detail and best left to your expertise. But either way, the common physics components should be synchronized.

I updated the Offline PR to include internal conversions randomly selected using the internal conversion branching fraction (measured for RPC). This now generates external and internal conversions at their expected rates with one tool entry in the capture products list, naturally keeping the spectra synchronized given any updates. I tested generating muon stop pileup to see that RMC is added at the expected rate (I saw 2 events with 20k generated stops) as well as that the internal conversions are generated at the expected rate (I increased the overall RMC rate for this test due to the low expected rate). Are there any other tests you would suggest running?

brownd1978 (approved):
As the underlying code now supports a 'physical' mode combining internal and external I will approve this, but expect updates once those changes are merged.

PR #361, opened 2024-11-09
Add dts filters associated with each dts mixing module

edcallaghan:
This PR adds a set of `MixingFilter`s to the standard mixing workflow, and is a sibling PR to https://github.com/Mu2e/Offline/pull/1367.

PR #369, opened 2024-11-25
IPA and DIO dts mixing filters

edcallaghan:
This PR implements extension of the configuration of the dts filters introduced for mixing workflows in https://github.com/Mu2e/Production/pull/361 via additional insertions into the "baseline fcl" from which a mixing dataset is produced. Overrides to the configuration are encapsulated in primary-conditioned fcl stubs located in the mixing directory. Appropriate implementations are provided for for IPA and DIO primaries. I've made a distinct PR because this deals with the interface for beginning production of a dataset as opposed to the fcl structure of mixing jobs, but https://github.com/Mu2e/Production/pull/361 should be merged first.


edcallaghan:
@michaelmackenzie tagging you for interest in implementing this for RMC products. `JobConfig/mixing/filters/DIOtail.fcl` contains the proposed configuration for DIO, so whatever this ends up as when merged could be a starting point.

JobConfig/mixing/filters/CeEndpoint.fcl
brownd1978, line 1: Can you please explain the intent of these empty files? Will they always be empty (null filtering) or will they eventually have content and these are just placeholders?

JobConfig/mixing/filters/DIOtail.fcl
```diff
@@ -0,0 +1,7 @@
+# remove steps descendant from stopping target DIOs generated in Mu2e code
+physics.producers.MuStopPileupFilter.selection: {
+  tool_type: ProcessVolumeDetectorStepAntiSelectionTool
+  processes: [ "mu2eMuonDecayAtRest" ]
+  volumes: [ "Foil_00", "Foil_01", "Foil_02", "Foil_03", "Foil_04", "Foil_05", "Foil_06", "Foil_07", "Foil_08", "Foil_09", "Foil_10", "Foil_11", "Foil_12", "Foil_13", "Foil_14", "Foil_15", "Foil_16", "Foil_17", "Foil_18", "Foil_19", "Foil_20", "Foil_21", "Foil_22", "Foil_23", "Foil_24", "Foil_25", "Foil_26", "Foil_27", "Foil_28", "Foil_29", "Foil_30", "Foil_31", "Foil_32", "Foil_33", "Foil_34", "Foil_35", "Foil_36" ]
```
brownd1978, line 5: Will these foils be relevant for all target-stop processes? If so, I suggest to define this sequence in prolog, then you can source it wherever needed.

JobConfig/mixing/filters/IPAMuminusMichel.fcl
```diff
@@ -0,0 +1,7 @@
+# remove steps descendant from IPA michels generated by G4
+physics.producers.MuStopPileupFilter.selection: {
+  tool_type: ProcessVolumeDetectorStepAntiSelectionTool
+  processes: [ "DIO" ]
+  volumes: [ "protonabs1" ]
+  momentum_threshold: 1.0
```
brownd1978, line 6: Is this momentum threshold value intended to be overridden in a real mixing job? If so I suggest to leave it blank, so that downstream jobs will fail unless it is set to a (correct) value. If you need a value for testing, you can set that in the test script.

Scripts/gen_Mix.sh
```diff
@@ -250,6 +250,8 @@ fi
 if [ "${PRIMARY}" == "NoPrimary" ]; then
   echo '#include "Production/JobConfig/mixing/NoPrimary.fcl"' >> mix.fcl
 fi
+# Override dts filters conditioned on primary
+echo "#include \"Production/JobConfig/mixing/filters/${PRIMARY}.fcl\"" >> mix.fcl
```
brownd1978, line 254: I see this is where you use the files defined above.
Since gen_Mix.sh is the 'master script' in the campaign I suggest to let it write a (transient) filter configuration directly into the main job fcl for the cases that needed it, and leave it out for the cases that don't. That avoids needing a large number of empty files.

brownd1978 (commented):
I suggest a slight modficiation to  avoid empty filter files, which also makes this design open to extension (for new primaries that don't overlap with pileup).

Scripts/gen_Mix.sh
edcallaghan: I'd argue for keeping a static-fcl-based workflow, even if constrained to obviate any empty files. I find all of the "master" campaign scripts to be a bit hard to parse, but the bulk of their output is constructed by stitching together upstream static fcl stubs. The current implementation keeps with that, so that a domain expert implementing / changing the filter for a given primary need only populate the corresponding fcl in `.../mixing/filters`. That does require that empty files exist for non-overlapping primaries (conversion electrons, etc.), but it relieves that implementer from having to touch or understand the campaign script.

The far alternative is to instead embed the actual filter implementations into a series of if-else clauses in the `gen_Mix.sh` itself, which is how I interpret your suggestion. The downside I see with that is that it will require continued interaction with / iteration of that script as more primaries are correctly implemented.

The middle ground would be either a) explicitly if-elseing the inclusion of filter fcl stubs conditioned on the primary, or b) extending the logic from "include the stub for this primary" to "include the stub for this primary if that stub exists", so that the empty files can be removed. (a) would require continued changes the script, and (b) allows for mistakes if someone doesn't test prudently, but we can probably survive that. What do you think of (b)?

JobConfig/mixing/filters/IPAMuminusMichel.fcl
edcallaghan: I think that this is the right place to define it. But it is good that you flagged it, because this needs to be updated to units of energy (already merged into `Offline`) and to match the selection of the primaries generated.

brownd1978:
gen_* scripts exist because fhicl is not a programming language (ie does
not support if-then-else). Bash on the other hand is a real language, and
we should avail ourselves of all its features, not restrict ourselves only
to what fhicl supports.
Requiring the specific primary filter fcl to exist for mixing requires
updating Production when a new primary is added. Instead putting the
intelligence in gen_Mix.sh allows running campaigns with new primaries
using an existing Production version.

As a hybrid you could define 'target_primary_filter.fcl' and
'ipa_stop_filter.fcl' to deal with those cases. That would eliminate empty
files and make the design open to extension. You would provide the energy
threshold and production code in the campaign config, which is probably the
best way to coordinate those.

JobConfig/mixing/filters/IPAMuminusMichel.fcl
sophiemiddleton: Do we always intend this to then be the same value? If not it would need to be overridden and setting this to nil to then be overridden in the gen_Mix might make sense.

JobConfig/mixing/filters/CeEndpoint.fcl
edcallaghan: As discussed below, these were meant to provide a working space for any forthcoming implementations. They are obviated and removed by the latest push, which only makes use of a stub if one is defined.

JobConfig/mixing/filters/DIOtail.fcl
edcallaghan: That is likely; this sequence is now defined in `.../JobConfig/mixing/prolog.fcl`.

JobConfig/mixing/filters/IPAMuminusMichel.fcl
edcallaghan: @sophiemiddleton these will be changed at exactly the frequency that the configuration of the primary dts jobs are changed. e.g., if were to generate DIO tail electrons with an 85 MeV threshold, then we'd need to configure the mixing job to filter out DIO pileup with an 85 MeV threshold.

Given that, I think we should follow a common pattern between the primary jobs and their corresponding mixing jobs. Because the former have their energy ranges ``hard-coded'' (in `Production/primary/{DIOtail,IPAMuminusMichel}.fcl`), I've hard-coded the same values here. Dave and I talked a bit about this earlier, and it doesn't seem like there is a great solution to keep them synchronized. My understanding is that the former will be overwritten by yet-higher-level logic when the campaign is actually run. If that is the case, then IMO we should override these in the same way, and the (almost) single point of maintenance is at the campaign level. What do you think?

sophiemiddleton: OK, sounds like its the best we can do for now

sophiemiddleton (approved):

edcallaghan:
Finally, @brownd1978, I did a few short tests comparing complete 2BB `NoPrimary` jobs with and without trivial filters in place. From those, the memory overhead of duplicating the `dts` collections is less than 4%.

brownd1978 (approved):
Looks good, thanks for making the requested changes.

PR #372, opened 2024-12-03
OffSpill now has separate central and loop helix paths/outputs

bonventre:


brownd1978 (approved):

brownd1978:
@FNALbuild run build test with Mu2e/Offline#1380

PR #373, opened 2024-12-04
Correct ipa dts filter configuration to act on MuBeamFlash stream

edcallaghan:
A typo slipped into https://github.com/Mu2e/Production/pull/369 by which the correct dts filtering for IPA Michel mixing jobs would be applied to the wrong input stream (`MuStopPileup`, instead of `MuBeamFlash`). This PR fixes that mistake.


brownd1978 (approved):
good catch

PR #386, opened 2025-01-15
Add trigger MC truth matching to digitization

brownd1978:


brownd1978:
@FNALbuild run build test

JobConfig/digitize/OnSpill.fcl
```diff
@@ -5,3 +5,4 @@ physics.producers.EWMProducer.SpillType : 1
 # configure 'Triggered' output to be signal
```
michaelmackenzie, line 5: The spill type is both set above and in the OnSpill_epilog.fcl, it can likely be removed from one of these places (perhaps from here).

JobConfig/digitize/OnSpill_epilog.fcl
```diff
@@ -0,0 +1,24 @@
+# set the event timing for OnSpill
+physics.producers.EWMProducer.SpillType : 1
+# this next produces additional hit-level payload to support MC truth matching.
+# It produces functionally equivalent output as the standard module but is slower, so don't use this script for timing studies
+physics.producers.TTmakeSH.FilterHits : false
+physics.producers.TTmakePH.TestFlag : true
+physics.producers.TTmakePH.StrawHitSelectionBits : ["EnergySelection","TimeSelection","RadiusSelection"]
+physics.producers.TTmakePH.StrawHitMask          : ["Dead"]
```
michaelmackenzie, line 8: Should these selection bits be defined within `Mu2e/mu2e_trig_config` to avoid needing to update multiple locations if this list changes ? Or are these the only bits that `TTmakeSH` can reasonably filter out?

JobConfig/digitize/OnSpill_epilog.fcl
```diff
@@ -0,0 +1,24 @@
...
+# add the trigger MC matching to the appropriate paths
+# apr
+physics.apr_highP : [@sequence::physics.apr_highP, TTAprKSFMC ]
```
michaelmackenzie, line 11: This is of course a point of by-hand maintenance as we discussed. `Mu2e/Production` will need to track when any new triggers of interest are added and if the `LoopHelixFit` instance changes in a given trigger path (e.g. if `apr_lowP_stopTarg_multitrk` changes to a new `LoopHelixFit` instance with lower momentum thresholds or different parameter constraints). I don't have any suggestions here though, just remarking on this point.

JobConfig/digitize/prolog.fcl
```diff
@@ -54,18 +50,40 @@ Digitize: {
...
   filters : {
-    # filter for Triggerable stream.  These should be events that might have
-    # triggered.  The default setting is a particle with enough tracker digis
+# filter for Triggerable stream.  These should be events that might have
+# triggered.  The default setting is a particle with enough tracker digis
     # in the right momentum range to be reconstructable as a helix
     # This filter should be overriden as needed for other kinds of primaries (straight cosmics, calo signals, ...)
     Triggerable : {
       module_type : StrawDigiMCFilter
       MinNDigis : 15
```
michaelmackenzie, line 84: Just as a separate question, is requiring 15 digis a bit strict? Could we expect some amount of 14 digi events to have pileup hits that lead to our signal or background process of interest triggering the event, which we would keep in downstream analysis?

JobConfig/digitize/prolog.fcl
```diff
@@ -156,15 +180,13 @@ Digitize: {
   #
   # trigger selections that can be used to finalize the triggerOutput selection, depending on digitization mode and source
   #
-  SignalTriggers : [ "tprDe_highP_stopTarg*", "cprDe_highP_stopTarg*" ]  # events with 'high' momentum tracks with good KinKal fits
+  SignalTriggers : [ "tpr*_highP*", "cpr*_highP*", "apr_highP*" ]  # events with 'high' momentum tracks with KinKal fits
```
michaelmackenzie, line 183: Should this include the MPR triggers as well? If not, should this also remove the TPR triggers as we've moved to APR as the default?

JobConfig/digitize/prolog.fcl
```diff
@@ -189,5 +211,8 @@ Digitize.Outputs : {
 # give paths specific numbers outside trigger path range
 Digitize.trigger_paths : ["0:DigitizePath", "1:TriggerablePath", "2:TriggerableCHPath" ]
 Digitize.EndPath : [ @sequence::Digitize.EndSequence, TriggeredOutput, TriggerableOutput ]
-
+# override the trigger digitization sequence
+TrigRecoSequences.artFragmentsGen : @local::Digitize.DigitizeSequence
```
michaelmackenzie, line 215: This additionally adds the CRV digi sequence, which I assume is fine but I'm just noting this change.

JobConfig/primary/prolog.fcl
```diff
@@ -153,16 +153,16 @@ Primary: {
...
+      module_type : GammaConversionPoints
+      g4ModuleLabel : "g4run"
+      doFilter : true
+      rMin : 0.   # Include target and IPA
+      rMax : 500.
+      zMin : 5000.
+      zMax : 8000.
+      eMin : 1.5
+      verbosity : 0
+      saveTree : true
```
michaelmackenzie, line 165: Is there a reason for turning back on this ROOT TTree output? My previous commit turned this off as it was only on for validation and I didn't intend to push it with it on.

Validation/ceDigi.fcl
```diff
@@ -1,4 +1,5 @@
 #include "Production/JobConfig/digitize/OnSpill.fcl"
+physics.producers.MakeSS : { module_type : NullProducer }
```
michaelmackenzie, line 2: I'm sure this is necessary, but out of curiosity why is this needed here but not for standard CE digi-making in JobConfig?

michaelmackenzie (commented):
This overall looks great to me and will be great to have this reco-MC matching for the relevant trigger paths. I've added a few comments, but most of these are questions for my own understanding and not suggestions.

JobConfig/digitize/OnSpill_epilog.fcl
brownd1978: Those bits are currently defined in TrkHitReco prolog. I might be able to steal them from there.

JobConfig/digitize/OnSpill.fcl
brownd1978: good catch, I'll update.

JobConfig/digitize/prolog.fcl
brownd1978: Keep in mind too that's the cut on the # of digis, the number of hits passing the preselection and time clustering will be less. IIRC these cuts correspond to >99% of tracks passing TrkQual>0.4.

JobConfig/digitize/prolog.fcl
brownd1978: Good point. I'll remove TPR.

JobConfig/digitize/prolog.fcl
brownd1978: CRV needs to be there, as otherwise the MC matching fails. It's a negligible cost, as it has to be called anyways on any event which triggers.

JobConfig/primary/prolog.fcl
brownd1978: This may have gotten lost in the merge conflict resolutions. Let me look...

Validation/ceDigi.fcl
brownd1978: It's because SurfaceSteps appeared late in MDC2020 so some samples don't have them. makeSS is a placeholder module for creating them if needed. There's no one right answer, this PR switches the default from creating them to not creating them. for MDC2025 we won't need makeSS outside of primary generation.

JobConfig/primary/prolog.fcl
brownd1978: This was flagged only as a change in indenting: the value of saveTree and all the related parameters are unchanged compared to your previous commit.

michaelmackenzie: Perhaps this is an issue of commit order? Here's the value at the HEAD: https://github.com/Mu2e/Production/blob/main/JobConfig/primary/prolog.fcl#L165
Could it be worth rebasing to check that the final value will be `false`?

michaelmackenzie (approved):

JobConfig/primary/prolog.fcl
michaelmackenzie: I see that you resolved this either way, thank you!

PR #392, opened 2025-01-27
Minor tweaks to mix and digi campaigns

oksuzian:


brownd1978 (approved):
Is this PR ready? Or do you still want to consolidate the reco output?

oksuzian:
> Is this PR ready? Or do you still want to consolidate the reco output?

I already did on gen_Reco.sh level.
I'm not happy with all these if statements, but I can't think of a better solution.

PR #393, opened 2025-01-31
Changes to S1 in the preparation for ExtractedCRY production 

oksuzian:


JobConfig/cosmic/ExtractedCRY.fcl
oksuzian, line 1: Hopefully, @ehrlich-uva can figure out how to include mother volumes for CRV to set the min ranges in CRV explicitly.
There are currently mother volumes:  CRSmother_CRV_E1, CRSmother_CRV_T1, CRSmother_CRV_T2.
However, they can't be referenced with minRangeRegionCuts as they were not constructed using the nestBox, nestTubs...

In the current implementation "minRangeCut: 10" has a small impact on CRV efficiency.
However, it would be more accurate to set it to a lower value in CRV.

brownd1978 (approved):

PR #396, opened 2025-02-08
Pass1 for CRV Wideband

ehrlich-uva:
Requires Offline PR 1413.


ehrlich-uva:
@brownd1978 We would like to set up an initial pass1 for the Wideband CRV data. This requires Offline PR 1413 and this PR. Could you please merge both of them and also create tags for Offline and Production? Thank you.

ehrlich-uva:
This PR does not interfere with any simulations or reconstructions.

Pass1/Wideband/CrvPedestal_widebandExtracted.fcl
brownd1978, line 1: what is wideband extracted?

brownd1978 (commented):
The contents look fine, I'm just confused by the name widebandExtracted.

Pass1/Wideband/CrvPedestal_widebandExtracted.fcl
ehrlich-uva: The geometry we are using here is similar to the geometry of the CRV at the extracted position, except that a different channel map is used and only a small number of channels are connected. Since this setup is located at Wideband where we also have other setups (e.g. wideband1module, wideband4modules), I called it widebandExtracted.

Pass1/ExtractedPosition/dummy.fcl
brownd1978, line 1: what purpose does this file serve?

ehrlich-uva: Github behaved weirdly with an empty directory. If I remember correctly, it didn't add this subdirectory. Since it turns out that I didn't even need this directory, I could delete this directory.

ehrlich-uva: I removed the file, which removed the entire directory.

brownd1978 (approved):

PR #398, opened 2025-02-16
Remove extra Mu2eKinKal prolog include

michaelmackenzie:
Remove an extra Mu2eKinKal prolog include line from the Validation reco fcl, as this throws an error as Offline/TrkPatRec/fcl/Particle.fcl is not yet included. This prolog is already included in JobConfig/reco/Reco.fcl so it is not needed here.


rlcee (approved):
Looks good, thanks

PR #402, opened 2025-02-21
Change default config to be processing dts with SurfaceSteps

brownd1978:
This PR modifies the digitization script to be compatible with primaries produced after MDC2020ap, where SurfaceSteps replaced StepPointMCs for passive materials. It also adds a script to patch primaries produced prior to that.


brownd1978:
@FNALbuild run build test

brownd1978:
@FNALbuild run build test

oksuzian (approved):

PR #405, opened 2025-02-28
collecting DAQ errors

ehrlich-uva:
This merge has to wait for the next release of artdaq_core_mu2e and Offline.


brownd1978:
Is this PR compatible with the p070 version of artdaq-core-mu2e?

ehrlich-uva:
Yes, it's compatible with p070. 
The full story is that the only significant change in this PR is to keep an additional data product in the output art file. This data product doesn't show up until the Offline PR 1424 is merged, which requires Offline PR 1431 (ENVSET 070).

rlcee (approved):

PR #406, opened 2025-02-28
spill splitter change

sophiemiddleton:


brownd1978 (approved):
Thanks for this, it's good prep for the next large production.

oksuzian (commented):
Sorry, I don't understand the purpose of this.
I would prefer to keep the signal stream all inclusive.

oksuzian:
> What study requires 30-250 MeV range, that can't run using the 'All' sample?

We'd like to have as much stats as possible for cosmic background estimation.
I don't think we plan to digitize the entire 'All' datasets.

Regardless, I don't understand what issue we're trying to solve.
We can mix Signal stream with 30-250 MeV range.

brownd1978:
After discussions with Yuri we decided this isn't needed for MDC2020 and something slightly different is needed for MDC2025. So I will close this PR.

PR #414, opened 2025-03-20
Increase MC filtering threshold on external RMC generation

michaelmackenzie:
Increase the MC filtering in the RMC external simulation to more closely match the generation energy threshold. As the photon energy range only goes down to 85 MeV, the electron/positron spectrum shape below this becomes unphysical due to the missing photons below 85 MeV. For an 80 MeV/c primary filter threshold, I found 12 events in a 100k simulation passed the primary particle filtering (~0.01-0.02%). For a 3 billion RMC photon generation (around the Run 1 expectation) this would lead to about 60k accepted events overall.

### DIFF
--- a/JobConfig/primary/RMCExternal.fcl
+++ b/JobConfig/primary/RMCExternal.fcl
@@ -4,6 +4,7 @@
 
 #include "Production/JobConfig/primary/RMC.fcl"
 
+physics.filters.PrimaryFilter.MinimumPartMom : 80.
 physics.producers.generate.decayProducts.mode : "external"
 physics.producers.FindMCPrimary.PrimaryProcess : "mu2eExternalRMC"
 outputs.PrimaryOutput.fileName : "dts.owner.RMCExternal.version.sequencer.art"


brownd1978 (commented):
Naively it seems 80 Mev as a photon cutoff is a bit high compared to the positron trigger threshold of 70 MeV. Have you tested this doesn't change bias the triggered distributions?

michaelmackenzie:
The photon generation starts at 85 MeV, so this sample can't be used for positron spectra below ~84.5 MeV/c. The closure approximation already predicts ~3 billion photons above 85 MeV during Run 1, so lowering this threshold any further makes it computationally difficult to generate a physical Run 1 spectrum. We could generate a shorter period (perhaps 1 week?) sample with a wider physical RMC spectrum for trigger studies, in which case we would need to adjust both this filter cutoff and the RMC generation energy range.

brownd1978:
For a physics study with higher thresholds (~95 MeV) this 80 MeV threshold seems fine. For a sample representing the triggered dataset I think it would need to be lower. I believe Sophie is planning to produce a small (~1 day) triggered ensemble dataset soon.

brownd1978 (approved):

PR #415, opened 2025-03-26
Update mixing configuration to follow Offline

brownd1978:
This follows Offline PR [1445](https://github.com/Mu2e/Offline/pull/1445)


brownd1978:
@FNALbuild run build test with Mu2e/Offline#1445

PR #416, opened 2025-03-31
Pawel plesniak/stm sim update

PawelPlesniak:
# STM simulation work
Defines the infrastructure used to run STM simulation studies, with details in doc-db [51487](https://mu2e-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=51487)

# Summary
## fcl files
`BeamToVD.fcl` - propagates the beam from EleBeamCat and MuBeamCat all the way down to the STM shielding house. Only keeps events if there is at least one particle travelling a chosen VD (defaults to 101, will be extended by other devs)
`BeamToVD1809.fcl` - fires 1809keV photons from muon target stop positions in TargetStopsCat. Also only keeps events if there is at least one particle travelling a chosen VD
`STMResampler.fcl` - propagates beam from a chosen VD to the STM detectors

## Shell scripts
`CountSTMMixedEvents.sh` - counts the number of events mixed (may need to go into `Offline` instead)
`gen_BeamToVD.sh` - generates the fcl files for BeamToVD.fcl
`gen_BeamToVD1809.sh` - generates the fcl files for BeamToVD1809.fcl
`gen_STMCat.sh` - generates the fcl files for concatenating the files generated with  STMResampler.fcl. Primarily used to 
remove data products that are no longer needed, and to simplify data handling
`gen_STMHPGeDigitization.sh` - generates the fcl files for HPGeReco.fcl
`gen_STMMix.sh` - generates the fcl files for Mix.fcl
`gen_STMMixedEventsCounter.sh` - generates the fcl files for CountMixedEvents.fcl


PawelPlesniak:
@AndrewEdmonds11 agreed to review this PR, @brownd1978 can he be added as the reviewer?

JobConfig/pileup/prolog.fcl
AndrewEdmonds11, line 1: The changes in this file probably want to be put into an STM-specific fhicl file. At the moment, this might break existing workflows

Scripts/gen_BeamToVD.sh
AndrewEdmonds11, line 1: This is a question I guess to the primary users of this repo. Are you OK with these scripts being directly in the Scripts/ directory, or would you prefer that they go into an STM subdirectory?

AndrewEdmonds11 (changes requested):
Thanks, Pawel! Just one important change request in the prolog.fcl, and then there's a more general question for the Production team

Scripts/gen_BeamToVD.sh
PawelPlesniak: I think this would be useful regardless, as we have this for the `JobConfig/pileup` dir. Amended in the last commit.

PawelPlesniak:
@FNALbuild run build test

AndrewEdmonds11 (approved):
Thanks for making the changes, Pawel. Approved

PawelPlesniak:
@brownd1978 Can this be merged?

PR #420, opened 2025-04-15
 Introduction of the lookup table

oksuzian:
Production workflow will use a lookup table to extract par file and local index to run a segment.
There will be one master campaign that runs (almost) all the jobs.
This simplifies the workflow and jobs monitoring.


brownd1978 (approved):
I can't make a detailed technical evaluation, but overall this looks good. I like the subdivision of the primary processes. I have a few questions:
- Do you expect to use these fils in the next (mdc2025) campaign? If so I would remove mdc2020 from the name. 
- is digireco also used for sim (POT, beam resampling, ...)? If so I suggest a more general name.
- digi and sim stages are pure MC, while reco can be run on either sim or real data. Do you expect Pass1 (PassN) reco campaign to be much different? Should it be split out of digireco?
- Is stage1 specific to cosmics? Or does it cover POT too?
- I don't see a pbar primary campaign, I guess because that isn't part of MDS. Will you need it eventually?

oksuzian:
> I can't make a detailed technical evaluation, but overall this looks good. I like the subdivision of the primary processes. I have a few questions:
> 
> * Do you expect to use these fils in the next (mdc2025) campaign? If so I would remove mdc2020 from the name.
Are you suggesting we rename (for example) mdc2020_prolog.cfg to prolog.cfg

> * is digireco also used for sim (POT, beam resampling, ...)? If so I suggest a more general name.
Yes, ultimately, there would be just a few campigns that drive the entire production. I have the following 4 campaigns in mind:

1. stage-1, resampling, filtering, merging, mixing - this will driven by index dataset, and the parameter file and index will be extacted through the lookup table. 
2. digi: 1->1 workflow driven by drainingn
3. reco: 1->1 workflow driven by drainingn
4. ntupling: 1->1 workflow driven by drainingn

> * digi and sim stages are pure MC, while reco can be run on either sim or real data. Do you expect Pass1 (PassN) reco campaign to be much different? Should it be split out of digireco?
We need to consolidate our scripts with Ray.
Conceptually, we follow similar 1->1 workflows, which are driven by drainingn algorythm in POMS.

> * Is stage1 specific to cosmics? Or does it cover POT too?
Stage-1 POT and beam should be similar. CORSIKA is a bit special, but can be generalized as well...I suppose

> * I don't see a pbar primary campaign, I guess because that isn't part of MDS. Will you need it eventually?
I have generated this:
dts.mu2e.PbarSTGun.MDC2020ap.art
described in Production/data/primary_pbar.json

PR #426, opened 2025-04-26
Set fastCloning to false by default in artcat and StripMC fcls

oksuzian:
Otherwise, it may result in an issue described in INC000001194060.

### DIFF
--- a/JobConfig/common/artcat.fcl
+++ b/JobConfig/common/artcat.fcl
@@ -23,6 +23,7 @@
 process_name: artcat
 source: { module_type: RootInput }
 services: { message: @local::default_message }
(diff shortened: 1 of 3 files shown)


brownd1978 (commented):
Is this a patch to deal with flawed or obsolete data? Artcat is used in general production, can you think of any ways this might cause problems later?

rlcee:
The message we see explains why fast cloning is automatically turned off in the example case, but we do not know why there is an error.  As long as a long-term investigation/solution is underway, then patches like this are justified, but should be flagged for removal when the fix is available.  Perhaps by a github issue.

brownd1978 (approved):

PR #427, opened 2025-05-02
Add production scripts for physical RPC spectra

michaelmackenzie:
Generate physical RPC spectra from infinite pion lifetime samples for use in MDS campaigns. This is in connection with https://github.com/Mu2e/Offline/pull/1479

### DIFF
--- a/JobConfig/primary/RPC.fcl
+++ b/JobConfig/primary/RPC.fcl
@@ -14,7 +14,7 @@ physics.producers.generate : {
   spectrum : {
     spectrumShape : Bistirlich
     elow : 50.0 #MeV
-    ehi : 139.5.
+    ehi : 139.5
     spectrumResolution : 0.1 #MeV
   }
   pionDecayOff : true # will apply surv prob
--- a/JobConfig/primary/RPCExternalPhysical.fcl
+++ b/JobConfig/primary/RPCExternalPhysical.fcl
@@ -1,0 +1,9 @@
+#
+# RPC External, physical pion time distribution
+#
+
+#include "Production/JobConfig/primary/RPCPhysical.fcl"
+
+physics.producers.generate.RPCType : "mu2eExternalRPC"
+physics.producers.FindMCPrimary.PrimaryProcess : "mu2eExternalRPC"
+outputs.PrimaryOutput.fileName: "dts.owner.RPCExternalPhysical.version.sequencer.art"
--- a/JobConfig/primary/RPCInternalPhysical.fcl
+++ b/JobConfig/primary/RPCInternalPhysical.fcl
@@ -1,0 +1,9 @@
+#
+# RPC Internal, physical pion time distribution
+#
+
+#include "Production/JobConfig/primary/RPCPhysical.fcl"
+
+physics.producers.generate.RPCType : "mu2eInternalRPC"
+physics.producers.FindMCPrimary.PrimaryProcess : "mu2eInternalRPC"
+outputs.PrimaryOutput.fileName: "dts.owner.RPCInternalPhysical.version.sequencer.art"
--- a/JobConfig/primary/RPCPhysical.fcl
+++ b/JobConfig/primary/RPCPhysical.fcl
@@ -1,0 +1,15 @@
+##
+# RPC photon spectrum, based on on Bistrilich spectrum
+# Use accept-reject method to produce a physical pion time distribution
+#
+
+#include "Production/JobConfig/primary/RPC.fcl"
+
+physics.producers.RPCAcceptReject : @erase
+physics.filters.RPCAcceptReject : {
+    module_type : WeightSamplingFilter
+    EventWeightModule   : generate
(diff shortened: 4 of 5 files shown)


sophiemiddleton (commented):
Thanks for helping implement this. I wonder the WeightScalingFactor - this has to be included in our MDS normalization scripts too, right?

michaelmackenzie:
I would ignore the weight scale factor in the MDS ensembling, and instead determine the number of RPC events as you currently are, just sampling from this dataset rather than from the unphysical dataset. If you need to determine the normalization from this sample directly, you should be able to get this by either by scaling N(events)/N(gen) by 1/weight scale or by using sum of weights / N(gen) I believe (sum of weights is printed at the end, without the scale factor). I will have to think a bit more though to be sure.

sophiemiddleton:
Ok, ill think it over too

brownd1978 (approved):
I'm unclear what's implemented behind this config, can you please describe this at the next sim/reco meeting? 

>  WeightScalingFactor : 10000. # Maximum time weight is e^-0 = 1, but (essentially) all RPC events have p(evt) < 1e-4, so increase sampling efficiency

sophiemiddleton:
Hi Dave, its just what i missed out from the RPC generator. I think we had something similar in MDC2018. Michael had the idea to implement it as a separate filter, thats all

michaelmackenzie:
Hi Dave,
> I'm unclear what's implemented behind this config, can you please describe this at the next sim/reco meeting?
> 
> > WeightScalingFactor : 10000. # Maximum time weight is e^-0 = 1, but (essentially) all RPC events have p(evt) < 1e-4, so increase sampling efficiency

I'm giving a talk tomorrow at the Developers meeting to discuss this update, but I can also show the slides at the Sim/Reco meeting if this is helpful. The scale factor here just adjusts the accept-reject input weights to be closer to 0-1, increasing the acceptance efficiency while maintaining the same distribution shape (as long as the maximum scaled weight is still less than 1).

michaelmackenzie:
If a scaled event weight is greater than 1 it will automatically be accepted since `weight > rand(0,1)`. This is true in general for accept-reject algorithms where you can over accept low weight events if you don't know the true maximum weight, under representing the clipped high weight region. The current settings add a printout for events of this case, but otherwise nothing is impacted downstream.

In the stopped pion datasets I saw 1 event with a high weight, but I haven't investigated it yet to understand why it has such a large weight. I can try to do this ahead of the meeting tomorrow though.

michaelmackenzie:
I think I found the source of the high weight event, a pion stop that doesn't appear to pass the 350 ns selection filter but appears in the stopped pion dataset anyway:
```
[RPCGun::MakeEventWeight] Evaluating the event weight from the particle proper time
  t0 =   76.7, t1 =  223.7, tp = 136, tau = 5.236
 Tau = 5.236, Weight = 0.00532
[RPCGun::addParticles] p_gamma = (-122.65,   -3.13,   10.05), t_gamma =  223.7
[WeightSamplingFilter::filter::RPCAcceptReject] Event 1201:0:1262: Scaled event weight is greater than 1 (weight = 53.207, base weight = 0.00532)
```
We could filter these events out, but I don't think it will be impactful since it will be a few extra events at t < 350 ns, well before the digitization threshold or any sensible track selection. Does this sound reasonable to you, Dave and Sophie?

brownd1978:
Throwing these out would bias the output.

 I suggest to keep a weight data product when producing these unweighted
events. Normally the weight would be one, but if the scaled weight exceeds
1 save that instead. That flags these events. You could even use the weight
in downstream analysis, I think it would be statistically correct.

David Brown            ***@***.***
Office Phone (510) 486-7261
Lawrence Berkeley National Lab
M/S 50R5008 (50-6026C)  Berkeley, CA 94720

michaelmackenzie:
I agree these shouldn't be filtered out by the accept/reject module, but they likely should be filtered out of the stopped pion dataset as they fail the selection used to create it (t > 350 ns). Perhaps this is an issue in the filtered output stream?

For the accept/reject module, I think it's a good idea to add an output weight that is 1 by default and greater than 1 in the case of events with `weight > max_weight` to correct for the poorly chosen `max_weight`. I can put in a PR for this this week. I'll have to think about the different input cases though to be sure it's generally correct (e.g. if `weight = constant = 1` but `max_weight = 0.5` --> 100% selection efficiency, do we want to have an event weight of `2` for all events?)

brownd1978:
A filtered sample with range of [0,.5] should have a weight of
max(weight,0.5).  That would give statically equivalent weight sum as
unfiltered.

David Brown            ***@***.***
Office Phone (510) 486-7261
Lawrence Berkeley National Lab
M/S 50R5008 (50-6026C)  Berkeley, CA 94720

michaelmackenzie:
Hi Dave, I've added this to PR https://github.com/Mu2e/Offline/pull/1483 and tested with toys it correctly handles these cases.

PR #435, opened 2025-06-05
Add a triggerable selection for two track events

michaelmackenzie:
Add an initial idea of a two-track "triggerable" MC selection to the digitization. This is off by default, where I've added specific fcl with this turned on for two track studies. This would target physics like photon conversions (RMC, RPC) and antiproton events. The threshold is initially very low as we study RMC conversions and how well we can trigger on events like coincident ~60 MeV/c + ~35 MeV/c electron-positron pairs. We will only generate these processes to begin with, but may turn this on in general as we turn on corresponding triggers.

### DIFF
--- a/JobConfig/digitize/OffSpill.fcl
+++ b/JobConfig/digitize/OffSpill.fcl
@@ -8,12 +8,12 @@ outputs.TriggeredOutput.SelectEvents : [
   @sequence::Digitize.TrkTriggers,
   @sequence::Digitize.CaloTriggers ]
 # Lower thresholds for LoopHelix triggerable
-physics.filters.Triggerable.MinParticleMom : 50.0
-physics.filters.Triggerable.MaxParticleMom : 300.0
+physics.filters.TriggerableLH.MinParticleMom : 50.0
+physics.filters.TriggerableLH.MaxParticleMom : 300.0
 # Define the default list of triggerable paths
-physics.TriggerablePaths : [ "TriggerablePath", "TriggerableCHPath"]
+physics.TriggerablePaths : [ "TriggerableLHPath", "TriggerableCHPath"]
 # Define the trigger path
-physics.trigger_paths : [ "0:DigitizePath", @sequence::physics.TriggerablePaths,  @sequence::Trig_physMenu.trigger_paths]
+physics.trigger_paths : [ "DigitizePath", @sequence::physics.TriggerablePaths,  @sequence::Trig_physMenu.trigger_paths]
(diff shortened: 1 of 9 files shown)


oksuzian (approved):

michaelmackenzie:
The plan is to investigate the RMC/RPC photon conversion, antiproton trigger efficiencies and the background-only trigger rates for potential two-track triggers (with the normal instantaneous rate vs. N(POT) and timing studies). Then this "triggerable" selection will be updated to reflect the trigger selection, and can be made turned on by default to reflect the trigger.

JobConfig/digitize/OffSpill.fcl
```diff
@@ -10,8 +10,11 @@ outputs.TriggeredOutput.SelectEvents : [
 # Lower thresholds for LoopHelix triggerable
 physics.filters.Triggerable.MinParticleMom : 50.0
 physics.filters.Triggerable.MaxParticleMom : 300.0
-physics.trigger_paths : [ "0:DigitizePath", "1:TriggerablePath" , "2:TriggerableCHPath" , @sequence::Trig_physMenu.trigger_paths]
-outputs.TriggerableOutput.SelectEvents : [ "TriggerablePath", "TriggerableCHPath" ]
+# Define the default list of triggerable paths
+physics.TriggerablePaths : [ "TriggerablePath", "TriggerableCHPath"]
+# Define the trigger path
+physics.trigger_paths : [ "0:DigitizePath", @sequence::physics.TriggerablePaths,  @sequence::Trig_physMenu.trigger_paths]
```
brownd1978, line 16: the 0:DigitizePath index is redundant and somewhat confusing now given TriggerablePaths are no longer explicitly indexed. Please remove the index.

brownd1978 (approved):
With a minor change I approve, now we've gotten confirmation mixed path declarations are OK. Note we must declare the unindexed paths before the the trigger paths when they are mixed, to avoid potential overlap between index and slot.

I am bothered by the asymmetry between TriggerablePath and TriggerableCHPath. I think we should replace TriggerablePath with TriggerableLHPath, to make the config clearer and more consistent. That's somewhat outside the scope of this PR, so if you want to skip that I will do it in a followon PR.

michaelmackenzie:
@FNALbuild run build test

JobConfig/digitize/OffSpill.fcl
michaelmackenzie: I removed this index for now, and we can add this back if we are able to update to have bits in the SelectEvents in the future.

PR #436, opened 2025-06-19
Updates to MDS scripts for MDS2

sophiemiddleton:
Contains latest scripts for MDS production

May require further edits but since a lot has changed I thought it best to PR

### DIFF
--- a/JobConfig/ensemble/python/make_template_fcl.py
+++ b/JobConfig/ensemble/python/make_template_fcl.py
@@ -210,7 +210,6 @@ if __name__ == "__main__":
     parser.add_argument("--BB", help="BB mode e.g. 1BB")
     parser.add_argument("--release", help="e.g. MDC2020ad")
     parser.add_argument("--livetime", help="simulated livetime")
-    #parser.add_argument("--rue", help="signal branching rate")
     parser.add_argument("--tmin", help="arrival time cut")
     parser.add_argument("--dioemin", help="min energy cut dio")
     parser.add_argument("--rpcemin", help="min energy cut rpc")


JobConfig/ensemble/scripts/Stage3_mixsignal.sh
```diff
@@ -0,0 +1,208 @@
+#!/usr/bin/bash
+usage() { echo "Usage: $0
+  e.g. Stage3_mixsignal.sh --known MDS2a --signal CeMLeadingLog --rmue 1e-13 --nexp 3
+  usage:
+  --owner = the username of your account (or mu2e if you are using mu2epro);
+  --known = known physics tag e.g. MDS2a
+  --rmue = chosen rmue e.g. 1e-14 (note this could be edited during the process so check print outs)
```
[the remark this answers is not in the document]
sophiemiddleton: Yes, rate is ok

JobConfig/ensemble/scripts/Stage3_mixsignal.sh
```diff
@@ -0,0 +1,208 @@
...
+echo "======= combined samples info =========">> ${KNOWN}.txt
+echo "signal= ${SIGNAL}">> ${KNOWN}.txt
+echo "Rmue= ${RMUE}">> ${KNOWN}.txt
+echo "livetime_combined= ${LIVETIME}">> ${KNOWN}.txt
+echo "npseudo_experiments= ${NEXP}">> ${KNOWN}.txt
+
+# build complete list
+rm filenames_All_${SIGNAL}
+rm filenames_All_${KNOWN}
+echo "looking for mcs.mu2e.${SIGNAL}OnSpillTriggered.${RELEASE}_${DBPURPOSE}_${DBVERSION}.art"
```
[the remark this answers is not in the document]
sophiemiddleton: Sure, we havent mixed the samples yet, but this should be the mixed samples, yes

JobConfig/ensemble/scripts/Stage3_mixsignal.sh
```diff
@@ -0,0 +1,208 @@
+#!/usr/bin/bash
+usage() { echo "Usage: $0
+  e.g. Stage3_mixsignal.sh --known MDS2a --signal CeMLeadingLog --rmue 1e-13 --nexp 3
```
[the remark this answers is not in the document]
sophiemiddleton: The stage refers to the stage in ensembling, it was explained in my talk 

S1= make the cinfig, configure primaries
S2= ensembling
S3= this 'mixing' with signal

I agree the term mix is not good though

JobConfig/ensemble/scripts/Stage3_mixsignal.sh
```diff
@@ -0,0 +1,208 @@
...
+  usage:
+  --owner = the username of your account (or mu2e if you are using mu2epro);
+  --known = known physics tag e.g. MDS2a
+  --rmue = chosen rmue e.g. 1e-14 (note this could be edited during the process so check print outs)
+  --signal = primary name of chosen signal e.g. CeMLeadingLog for the e- ce leadinglog samples
+  --release = SimJob tag e.g. MDC2020aw
+  --dbpurpose = db purpose of input mcs files e.g. perfect or best
+  --dbversion = db version e.g. v1_3
+  --nexp = number of sets of mixed samples or 'pseudo experiments' to make default is 1
+  --chosenlivetime = chose a livetime in seconds e.g 86000
```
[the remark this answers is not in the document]
sophiemiddleton: The idea is that the user may not want the full sample, so they csn chose some smaller value. But the script rounds to the closest number of full files and tells them what the actual livetime used is.

JobConfig/ensemble/scripts/Stage2_submitensemble.sh
```diff
@@ -52,20 +52,19 @@ done
 NJOBS="" #to help calculate the number of events per job
 LIVETIME="" #seconds
 RUN=1201
-DIO_EMIN=""
+DIO_EMIN=95
 RPC_EMIN=""
```
brownd1978, line 56: Will there ever be a case where EMIN is different for different components? It would seem hard to analyze in that case. If these will always be the same I suggest a global EMIN value.

JobConfig/ensemble/scripts/Stage2_submitensemble.sh
```diff
@@ -52,20 +52,19 @@ done
...
 SAMPLINGSEED=1
-COSMICTAG=""
+COSMICTAG="MDC2020ar"
 GEN=""
 CONFIG=${TAG}.txt
 OUTRELEASE="MDC2020"
-OUTVERSION="ar"
-
+OUTVERSION="at"
+SURV=0.000891923
```
brownd1978, line 67: where does this number come from?

JobConfig/ensemble/python/normalizations.py
```diff
@@ -283,8 +285,13 @@ def ipaMichel_normalization(onspilltime, ipa_demin, run_mode = '1BB'):
         print("IPA_fraction_sampled=", fraction_sampled)
         nIPA = POT * ipa_stopped_mu_per_POT * IPA_decays_per_stopped_muon * fraction_sampled
         return nIPA
-    
 
+# work from signal to rmue  
+def get_ce_rmue(onspilltime, nsig, run_mode = '1BB'):
+    POT = getPOT(onspilltime, run_mode)
+    captures_per_stopped_muon = 0.609 # for Al
```
brownd1978, line 292: I appreciate there are a lot of numbers needed for these calculations. Would it be possible to declare them all at once at the top of the script? That would make it easier to maintain.

JobConfig/ensemble/scripts/Stage2_submitensemble.sh
sophiemiddleton: they are different

JobConfig/ensemble/fcl/prolog.fcl
```diff
@@ -6,7 +6,7 @@ Ensemble : {
           "keep mu2e::CosmicLivetime_*_*_*",
           "drop *_genCounter_*_*",
           "drop *_protonBunchIntensity_*_*"]
-    OutputCommandsMC : [   "keep *_*_*_*"]
+    OutputCommandsMC : [   "keep *_*_*_*","keep mu2e::SurfaceSteps_*_*_*"]
```
brownd1978, line 9: The additional 'keep' is redundant with the universal keep.

Scripts/CreatePionEfficency.sh
```diff
@@ -3,6 +3,15 @@
...
-sed -i -e 's/dts\.mu2e\.//' -e 's/sim\.mu2e\.//' -e 's/\..*\.art//' -e 's/ IOV//' $1_SimEff.txt
+mu2eGenFilterEff --out=$1_SimEff.txt --chunksize=100 --firstLine "PionSimEfficiencies" --verbosity 3 sim.mu2e.PiBeamCat.$1.art sim.mu2e.PiTargetStops.$1.art 
+
+#for  filter version you need the eventCount of sim.mu2e.PiTargetStops.$1.art as the denominator
+PREFILTER=$(samDatasetsSummary.sh sim.mu2e.PiTargetStops.$1.art  | awk '/Triggered/ {print $2}')
+FILTER=$(samDatasetsSummary.sh sim.mu2e.PiTargetFilt.$1.art  | awk '/Triggered/ {print $2}')
+FILTEREFF=$(awk "BEGIN {printf \"%.5f\", ${FILTER}/${PREFILTER}}") 
+echo "PiTargetFilt, ${PREFILTER}, ${FILTER}, ${FILTEREFF}" >> $1_SimEff.txt
+
+TOTALWEIGHT=37468 #TODO need better way of adding to the file
```
brownd1978, line 14: What is the point of these variables?  They seem to have no effect on the script.

JobConfig/ensemble/python/normalizations.py
```diff
@@ -174,13 +174,14 @@ def corsika_offspill_normalization(livetime, run_mode = '1BB'):
...
-  selected_sum_of_weights = 0.178864 # from filter - TODO
-
+  npistops = 41798702 ## from above (MDC2020aw)
+  target_stopped_pi_per_POT =  0.003924801 * 0.522483775 # from above (MDC2020ak)
+  npifilter = 4378518 # from above (MDC2020aw)
+  time_eff = npifilter/npistops  
+  total_sum_of_weights =  37468# from filter
+  selected_sum_of_weights =  5.7055#from filter
+
+  gen_factor = 50 # from Michael (for the new sampling factors)
```
brownd1978, line 184: Can you add to the comment how these numbers were extracted? Ie > samweb dh.dataset="....", etc
We need a long-term plan how these numbers can be determined from production output directly.

brownd1978 (commented):
Further feedback

Scripts/CreatePionEfficency.sh
sophiemiddleton: This is in flux, we dont know what will be needed once Michael finishes his work

sophiemiddleton:
@brownd1978 suggests switching the .fcl to be an ntuple

We also want to think about using nevts to sample fractions of files

sophiemiddleton:
@brownd1978 I have fixed most of the issues.

I think there will be further enhancements of the signal script and normalizations to tidy things up but I think as an update to what is in the HEAD this is hopefully a good step forward

sophiemiddleton:
@brownd1978 can we merge this now so we can get some of the more important changes.

I think for final signal script, let's wait until we have the new samples and the mixed versions.

PR #441, opened 2025-07-17
Reorganization of RPC generation and adding physical pion stop sampling

michaelmackenzie:
This is in connection to the Offline RPC updates for the physical pion sampling. Pion stops are now selected using the `StopSelection` module, so the pion stop is consistent with the filters applied (e.g. the timing cut). The only thing I'm not sure about is where to put the physical pion stop dataset creation fcl, as it seems more part of the "beam" jobs, but I want to include the "primary" prolog to keep the RPC `tmin` cut the same and also get the resampling setup from there to not repeat logic. Should I move this to "primary" or keep it in "beam"?


sophiemiddleton (approved):
Looks good to me. Thanks Michael for doing this!

PR #442, opened 2025-07-18
Various updates to MC production scripts

oksuzian:


Scripts/convert_to_exp.sh
```diff
@@ -0,0 +1,29 @@
...
+        print
+    }'
+}
+
+# Read from file if provided, otherwise from stdin
+if [ $# -eq 1 ]; then
+    convert_numbers < "$1"
+else
+    convert_numbers
+fi
```
brownd1978, line 29: Looks like a blank line.

brownd1978 (approved):
I can't provide a technical review.  I note there are a lot of hard-coded values that will need to be hand-curated, but I don't see an alternative.

PR #458, opened 2025-09-23
New signal ensembling scripts and tutorial material for users

sophiemiddleton:
Contains:
* New scripts
* markdown containing tutorial on how to use
* a few minor fixes to python
* split.fcl tempalte (used by the new scripts)

### DIFF
--- a/JobConfig/ensemble/fcl/split.fcl
+++ b/JobConfig/ensemble/fcl/split.fcl
@@ -2,13 +2,11 @@
 
 process_name: artsplit
 
-# read only the first 100 events
+# read only the first n events
 source: {
   module_type: RootInput
   fileNames: @nil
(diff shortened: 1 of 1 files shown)


JobConfig/ensemble/MakeSignalMDS_tutorial.md
```diff
@@ -0,0 +1,52 @@
+# Introduction
```
brownd1978, line 1: Is there a reason this belongs here instead of Tutorial repo?

sophiemiddleton: I think we need to discuss that more generally.

I will check witj Andy. Recently weve pit other tutorials in EventNtuple too, so we need a global plan

brownd1978: OK. For now lets go with this. Let me know when you and Andy have a global plan.

PR #464, opened 2025-10-07
Create extracted for legacy

brownd1978:
Explicitly specify extracted config. Change nesting so this is defined in 1 place (cosmic/geom_cosmic_extracted.txt)


brownd1978:
@FNALbuild run build test with Mu2e/Offline#1615

brownd1978:
@FNALbuild run build test with Mu2e/Offline#1615

PR #477, opened 2025-11-07
Fixes for MDC2025 mixing

brownd1978:
These fixes allow creating a functional mixing job. Some configurations are specific to how the MDC2025 mixin collections were stored. The functional job also requires an updated SimEfficiencies table which isn't yet loaded into the database.


Scripts/gen_Mix.sh
```diff
@@ -39,14 +40,16 @@ PBEAM=""
 EARLY=""
 MERGE_EVENTS=5000 #source events in mixing
 OWNER=mu2e
-NEUTNMIXIN=50
+NEUTNMIXIN=1
```
oksuzian: Why did it change to 1?

brownd1978:
David BrownOn Nov 7, 2025, at 09:32, oksuzian ***@***.***> wrote:﻿
@oksuzian commented on this pull request.



In Scripts/gen_Mix.sh:
> @@ -39,14 +40,16 @@ PBEAM=""
 EARLY=""
 MERGE_EVENTS=5000 #source events in mixing
 OWNER=mu2e
-NEUTNMIXIN=50
+NEUTNMIXIN=1

Why did it change to 1?I think you merged it more.  There are only 40 files.  We should run a test mixing and check for oversampling. I can do that later today.  



On Scripts/CreateSimEfficiency.sh:
Why hard-coded values?Because the input now is split across several musings. It makes no sense to have so many parameters.

oksuzian (approved):

PR #510, opened 2026-02-16
Remove database default from JobConfig, add scripts

bonventre:
After this PR instead of running a JobConfig fcl locally out of the box users should use addDatabase.sh to create fcl with explicit database settings first

### DIFF
--- a/JobConfig/common/database_empty.fcl
+++ b/JobConfig/common/database_empty.fcl
@@ -1,0 +1,2 @@
+services.DbService.purpose: EMPTY
+services.DbService.version: ""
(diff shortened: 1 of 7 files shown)


rlcee:
If one includes
fcl/standardServices.fcl
they get "EMPTY" purpose so adding an empty epilog doesn't change the setting.
Did you want the epilogs explicitly as a visual placeholder?

bonventre:
Checking copilot comments
1. Nothing outside of validation uses database.fcl
2. will need to retrigger tests
3. added description
4. scripts is in PATH so removed relative path
5/6. added error handling
7. executable permissions are added in git, not sure why copilot doesn't see
8. fixed
9. added
10. fixed

rlcee (approved):

PR #512, opened 2026-02-17
Add output product

brownd1978:
Add output product to hand-curated validation list


rlcee (approved):
Sorry I missed this - thanks

PR #534, opened 2026-04-19
Update to MDS workflows

sophiemiddleton:
This pull request updates the documentation for the ensemble mock data pipeline to better reflect the current workflow and script naming conventions, clarify the multi-stage process, and ensure consistency in parameter descriptions and output formats. The changes improve clarity for users running the pipeline and align documentation with recent script and process updates.

**Documentation and Workflow Updates:**

* Updated the main `README.md` to clarify each stage of the pipeline, map documentation files to pipeline stages, and provide a clearer overview of the workflow.

**Script Naming and Usage Consistency:**

* Renamed scripts throughout documentation to reflect new names and stage numbers:
  - Stage 1: `Stage1_makeinputs.sh` → `Stage1_initate_ensemble.sh` [[1]](diffhunk://#diff-32703c6dbb56edce8c9b33c3a109cbca0a945eadc3eab6e7bc94ab4cc2135ef0L1-R1) [[2]](diffhunk://#diff-32703c6dbb56edce8c9b33c3a109cbca0a945eadc3eab6e7bc94ab4cc2135ef0L11-R18)
  - Stage 2: `Stage2_submitensemble.sh` → `Stage2_build_sampler.sh` [[1]](diffhunk://#diff-308edfc2276c525f254762d263a197a01758083c2e2a692835409426eb4a7fc4L1-R1) [[2]](diffhunk://#diff-308edfc2276c525f254762d263a197a01758083c2e2a692835409426eb4a7fc4L10-R33)
  - Stage 4: `Stage3_addsignal.sh` → `Stage4_signal.sh` [[1]](diffhunk://#diff-86bb5c1f4f9e6d36b5817c0b7ad37b0c868119a72b141f4aabd7058f82c06874L1-R18) [[2]](diffhunk://#diff-86bb5c1f4f9e6d36b5817c0b7ad37b0c868119a72b141f4aabd7058f82c06874L45-R45)
  - Stage 6: `combine_ntuples.sh` → `Stage6_blind_and_mix.sh` [[1]](diffhunk://#diff-59889268f2f3a109239deb23a22f2e3e5f8e6ad6edfb9bba288edb80aa3d0c97L30-R30) [[2]](diffhunk://#diff-59889268f2f3a109239deb23a22f2e3e5f8e6ad6edfb9bba288edb80aa3d0c97L51-R58) [[3]](diffhunk://#diff-86bb5c1f4f9e6d36b5817c0b7ad37b0c868119a72b141f4aabd7058f82c06874L86-R95) [[4]](diffhunk://#diff-86bb5c1f4f9e6d36b5817c0b7ad37b0c868119a72b141f4aabd7058f82c06874L104-R104)

**Parameter and Output Format Improvements:**

* Updated default parameter values (e.g., release/version tags, generator names) and clarified descriptions in Stage 1 and 2 documentation for better accuracy and alignment with current practices. [[1]](diffhunk://#diff-32703c6dbb56edce8c9b33c3a109cbca0a945eadc3eab6e7bc94ab4cc2135ef0L35-R37) [[2]](diffhunk://#diff-308edfc2276c525f254762d263a197a01758083c2e2a692835409426eb4a7fc4L10-R33)
* Standardized output files to be in shell-sourceable format (key="value" pairs) and clarified that all parameters and event yields are written in this format, simplifying downstream usage. [[1]](diffhunk://#diff-32703c6dbb56edce8c9b33c3a109cbca0a945eadc3eab6e7bc94ab4cc2135ef0L52-R52) [[2]](diffhunk://#diff-32703c6dbb56edce8c9b33c3a109cbca0a945eadc3eab6e7bc94ab4cc2135ef0L65-R65) [[3]](diffhunk://#diff-308edfc2276c525f254762d263a197a01758083c2e2a692835409426eb4a7fc4L45-R52)

**Stage and Process Clarifications:**

* Clarified the sequence and dependencies between stages, including which configuration files are produced and consumed at each step, and updated references to reflect new script and file names. [[1]](diffhunk://#diff-308edfc2276c525f254762d263a197a01758083c2e2a692835409426eb4a7fc4L45-R52) [[2]](diffhunk://#diff-86bb5c1f4f9e6d36b5817c0b7ad37b0c868119a72b141f4aabd7058f82c06874L45-R45)
* Updated references to which stages use certain functions (e.g., Poisson sampling now noted as used in Stage 4, not Stage 3).

These changes collectively make the documentation more accurate, user-friendly, and maintainable.

** Note on Development: **

* Stages 1-3 have been tested final stage testing still in progress, although they were tested in previous iterations
* CoPilot used to tidy up some of the text out puts in the .sh scripts to make them more instructive and user friendly
* Additional work in coming months to enhance automation.


oksuzian:
# Code Review: PR #534 – Update to MDS workflows

Below are the issues I found, grouped by severity. Overall the refactor (shell-sourceable configs, validation checks, constants module, nicer output) is a nice direction, but there are several real bugs and some copy-paste accidents that should be fixed before merging.

---

## 🔴 Bugs (must fix)

### 1. Syntax error in `test_normalizations.py` – duplicate kwarg
```python name=JobConfig/ensemble/python/test_normalizations.py url=https://github.com/Mu2e/Production/pull/534/files#diff-test_normalizations
pot_2bb_spill = normalizations.get_pot(
    self.on_spill_time_2BB, run_mode='2BB',
    printout=False, printout=False, method='spill'
)
```
`printout=False` is passed twice → `SyntaxError`. The test file cannot even be imported/run.

### 2. `Stage4_signal.sh` reads a non‑existent file
```bash
while IFS= read -r file_path; do
    if [ -f "$file_path" ]; then
        cp "$file_path" ${KNOWN}.txt
    fi
done < config.txt      # <-- config.txt is never created; the line that used to create it is commented out
```
`config.txt` is not produced anywhere in this script (the `mu2eDatasetFileList ... >> config.txt` line above it is commented out). The loop silently becomes a no‑op and `${KNOWN}.txt` may not exist or may be stale. Same issue is duplicated in `Stage5_signal.sh`.

### 3. Hard‑coded `MDS3a` in `Stage4_signal.sh` / `Stage5_signal.sh`
```bash
N_TOTAL_KNOWN=$(samDatasetsSummary.sh mcs.${OWNER}.ensembleMDS3aMix1BBTriggered.${RELEASE}_${DBPURPOSE}_${DBVERSION}.art | ...)
...
mu2eDatasetFileList nts.mu2e.ensembleMDS3aMix1BBTriggered.${EVENTNTUPLE}.root > filenames_All_${KNOWN}
```
`MDS3a` is hard‑coded where `${KNOWN}` should be used. So running with `--known MDS2c` will still query the `MDS3a` dataset. The BB mode `1BB` is also hard‑coded instead of `${BB}`.

### 4. `--signal` argument is ignored in `Stage4_signal.sh`
The old script used `--prc ${SIGNAL}`; the new one is:
```bash
NSIG=$(calculateEvents.py --livetime ${LIVETIME} --prc "CeMLeadingLog" --BB ${BB} --rue ${RATE})
```
So the user‑supplied `--signal` is no longer used in the yield calculation. Anything other than `CeMLeadingLog` will silently be computed wrong.

### 5. `Stage5_signal.sh` is an accidental duplicate of `Stage4_signal.sh`
Both files:
- Start with `echo "Stage 4: Signal Processing"` (Stage5 labels itself as Stage 4).
- Have the same usage line `Stage4__signal.sh ...` (with double underscore).
- Are not referenced in any documentation.

`Stage5_signal.sh` should be removed, or if it is genuinely intended to be a separate stage its content needs to be different and documented.

### 6. `BEAM_POT` in `Stage1_initate_ensemble.sh` will always be empty
```bash
BEAM_POT=$(echo "${BEAM_INFO}" | grep "^POT=" | awk '{print $NF}')
```
`normalizations.get_pot(...)` never prints a line beginning with `POT=`; it prints `NPOT=`, `NMOT=`, `Tcycle=`, `livetime=`, `on_spill_time=`. So `beam_pot=""` always ends up in `${TAG}.txt`. Either print a `POT=` line in `get_pot`, or grep for `^NPOT=`.

### 7. Typo in script name: `Stage1_initate_ensemble.sh` (should be `initiate`)
```bash name=JobConfig/ensemble/scripts/Stage1_initate_ensemble.sh url=https://github.com/Mu2e/Production/pull/534/files#diff-stage1
Stage1_initate_ensemble.sh
```
This typo is propagated everywhere (README, BuildingConfiguration.md, BuildingKnownPhysics.md, UploadToTape.md, error messages). Rename to `Stage1_initiate_ensemble.sh` while it is still easy to fix — downstream users will inherit the typo otherwise.

### 8. `Stage2_build_sampler.sh`: self‑assignment no‑ops, plus broken variable mapping
```bash
if [[ ! -z ${TMIN} ]]; then
  TMIN=${TMIN}          # no-op
fi
if [[ ! -z ${RMC_kmax} ]]; then
  RMC_kmax=${RMC_kmax}  # no-op
fi
```
These blocks do nothing. Also note the Stage‑1 output uses capitalized `TMIN=` and `RMC_kmax=` (mixed case). The mapping for `TMIN` is OK by coincidence, but style‑wise the block is misleading. More importantly, `Stage1_initate_ensemble.sh` writes `TMIN="..."` not `tmin="..."`, and Stage 2 then relies on it — double‑check that the casing matches across both scripts (`BB`, `TMIN`, `RMC_kmax` are capitalized; `onspilltime`, `njobs`, `dio_events`, etc. are lower). It's easy to get bitten later.

### 9. `UploadToTape.md` doc contradicts the script
- Doc says the script uploads **to tape**: `mu2eFileUpload --tape`.
- Script actually calls: `UPLOAD_OUTPUT=$(ls *.tar | mu2eFileUpload --disk 2>&1)` and registers with `enstore:` prefix.
- Doc's default `--owner` is `sophie`; script default is `mu2e`.
- Doc says the script name is `Stage3_configure_ensemble_campaign.sh` but then: *"Note: This is actually Stage 2b in the pipeline"* — pick one. If it's 2b, don't name it Stage3.
- Doc shows `samweb add-file-location ${TAR_FILE} ${TAPE_PATH}` but the script uses `cnf.${OWNER}.ensemble${TAG}.${RELEASE}${VERSION}.tar` (no `.0`) while the actual file is `...0.tar` — that's likely a SAM registration bug too.

### 10. Duplicate import in `normalizations.py`
```python
from constants import (
    ...
    ONEBB_CYCLE,
    ...
    ONEBB_CYCLE,   # duplicated
    ...
)
```
`ONEBB_CYCLE` appears twice in the import list.

### 11. `calculateEvents.py`: double `parse_args()`
```python
args = parser.parse_args()
(args) = parser.parse_args()
main(args)
```
Second call is pre‑existing but worth cleaning up while you're here.

### 12. `constants.py` missing newline at EOF
Minor, but `\ No newline at end of file` — please add.

---

## 🟡 Logic / design concerns

### 13. `get_pot(..., method='spill')` changes the meaning of POT
Old: POT = `num_cycles * pot_per_cycle` (uses duty factor implicitly through cycle/livetime relationship).
New default: POT = `on_spill_time / SPILL * protons_per_spill`.

These two methods differ by ~duty-factor-level amounts; the test asserts they agree within 5%, but the downstream `ce_normalization`, `dio_normalization`, etc. all now use `method='spill'` (the new default) while all prior tuning/expected yields were done with the cycle method. Worth confirming this intentional shift with the physics group, and flagging it in a changelog.

### 14. `get_pot` parameter `frac=False`
Default was `frac=1`. Now it is `frac=False` (a bool) and `frac` is no longer referenced in the function body. If `custom` mode is still desired, the logic is gone; if it isn't, remove the argument.

### 15. `ipa_stopped_mu_per_POT` print removed, but the setup loop's control flow was odd
Before, the `print("IPAStopMuonRate=", ...)` line was indented incorrectly (module-level), now commented out. The preceding loop still silently multiplies `ipa_stopping_rate` by every `float(words[3])` it finds — if the loop matches multiple rows this compounds. Worth a review for correctness.

### 16. `Stage2_build_sampler.sh`: dataset validation uses integer arithmetic on scientific‑notation yields
```bash
expected_yield_var=$(printf "%.0f" "$expected_yield_raw" 2>/dev/null)
...
if (( generated_events < expected_yield_var )); then
```
This is fine for ordinary numbers, but many of the config values are scientific notation (e.g. `1.42e+07`) and `(( ... ))` doesn't handle that without the `printf` conversion — which you do — good. However `generated_events=$((generated_events + 0))` earlier discards fractional POT values. Ensure the `samDatasetsSummary.sh` output is truly an integer, otherwise switch to `awk` comparisons.

### 17. `Stage3_configure_ensemble_campaign.sh`: `--append` path defines `njobs` from nothing
```bash
njobs=${njobs:-100}
```
In append mode the config file isn't sourced; `njobs` comes from the environment. If it isn't set, the campaign JSON silently uses 100. Either source the config file in append mode too, or require `--tag` to be paired with a readable `${TAG}.txt`.

### 18. Stage2 command prints without quoting
```bash
cmd="mu2ejobdef --desc=ensemble${TAG} ..."
$cmd
```
Building a command as a string and re-executing via `$cmd` breaks on arguments containing spaces or when arguments have colons/commas that the shell might re-split. Prefer a bash array:
```bash
cmd=( mu2ejobdef --desc=ensemble${TAG} ... )
"${cmd[@]}"
```

### 19. `Stage2_build_sampler.sh` no longer submits jobs
The old `Stage2_submitensemble.sh` ended with `mu2ejobsub …`. The new script stops at `mu2ejobdef` + index definition and the documentation still says *"A multi-component simulation job is queued on the grid, ready to process events."* That's no longer true — it's only the jobdef + SAM idx that are created; actual submission now happens in POMS (per UploadToTape.md). Update the doc, and in particular update BuildingKnownPhysics.md §1 which claims submission happens here.

### 20. `Stage6_blind_and_mix.sh` still titled "Stage 6" but the banner says "Stage 6: Blind and Mix" while file only does `hadd` merging (no real "blinding"). Consider renaming or adding the actual blinding step; otherwise the name is misleading.

---

## 🟢 Minor / nits

- `Stage4_signal.sh` usage message shows `Stage4__signal.sh` (double underscore).
- `Stage4_signal.sh` first `echo` banner says *"Stage 4: Signal Processing"* — Stage5 says the same (reinforces that Stage5 is a dup).
- `BuildingConfiguration.md` says `--gen` default is `CRY`, but the script's default is `Signal` (`GEN="Signal"`). These disagree.
- `BuildingConfiguration.md` describes `CosmicGen` value `CRY or CORSIKA`, but Stage 1 builds dataset as `dts.mu2e.Cosmic${GEN}.${COSMICS}.art` → with default `GEN=Signal` that yields `dts.mu2e.CosmicSignal...`, which is what `Stage2_build_sampler.sh` also expects (`dts.mu2e.CosmicSignal.${COSMICTAG}.art`). So the docs and defaults disagree with the actual dataset naming convention.
- Normalizations: all the `if VERBOSE: print(...)` changes are fine, but removing the unconditional `print("IPAStopMuonRate=", ...)` means startup is silent; consider leaving it gated behind `VERBOSE`.
- `logchecker.py`: nice aggregation addition; consider printing totals sorted, and adding a `-o/--output` flag instead of hard-coded file names.
- `Stage1_initate_ensemble.sh`: `rm -f ${COSMICS}` where `COSMICS=MDC2025ac` — if a user ever runs this in a directory containing a directory with that name you'd error out (not deletable by `rm -f`). Benign, but consider using a more specific filename like `filenames_${COSMICS}.txt`.

---

## Summary of top things to fix before merge

1. Fix `test_normalizations.py` duplicate kwarg (**syntax error**).
2. Fix hard-coded `MDS3a` / `1BB` / ignored `--signal` in `Stage4_signal.sh` and remove `Stage5_signal.sh` if unintended.
3. Fix `config.txt` read in `Stage4_signal.sh` (the file is never created).
4. Fix `BEAM_POT=` empty capture in `Stage1_initate_ensemble.sh`.
5. Rename `Stage1_initate_ensemble.sh` → `Stage1_initiate_ensemble.sh` and update all references.
6. Remove duplicate `ONEBB_CYCLE` import in `normalizations.py`.
7. Reconcile `UploadToTape.md` with the actual script (tape vs disk, owner, stage number vs script name, `TAR_FILE` registration path).
8. Update Stage 2 docs — script no longer submits to the grid.
9. Remove dead self‑assignment blocks in `Stage2_build_sampler.sh` and audit variable-name casing between Stage 1 and Stage 2.
10. Decide whether the new default `method='spill'` in `get_pot` is intended (it shifts yields).

Happy to help draft any of these fixes if you'd like — just say the word and I can open a follow‑up PR on your branch.

sophiemiddleton:
I will add in the AI suggestions for everything except Stages > 4 as these are still works in progress. I have not been able to test the new scripts for those stages due to an EventNTuple bottle neck which is being worked on by Andy and I (basically we need a new Analysis Musing with several new PRs merged into it)

PR #535, opened 2026-04-21
Remove duplicate SelectReco calls

brownd1978:
Cleanup to remove unused definition (OffSpillPath) and redundant calls to SelectReco in all the reco paths. Those calls were innocuous as art only invokes a module once, but having them was confusing.

### DIFF
--- a/JobConfig/reco/prolog.fcl
+++ b/JobConfig/reco/prolog.fcl
@@ -365,21 +365,16 @@ Reconstruction.OnSpillPath : [
   SelectReco
 ]
 
-# OffSpill paths. Separate by fit type
-Reconstruction.OffSpillPath : [
+Reconstruction.CentralHelixPath : [
   @sequence::Reconstruction.PreTrk,
   @sequence::Reconstruction.OffSpillTrk,
-  SelectReco
-]
-
-Reconstruction.CentralHelixPath : [
-  @sequence::Reconstruction.OffSpillPath,
   CHFilter,
   SelectReco
 ]
 
 Reconstruction.LoopHelixPath : [
-  @sequence::Reconstruction.OffSpillPath,
+  @sequence::Reconstruction.PreTrk,
+  @sequence::Reconstruction.OffSpillTrk,
   LHFilter,
   SelectReco
 ]
--- a/JobConfig/recoMC/prolog.fcl
+++ b/JobConfig/recoMC/prolog.fcl
@@ -67,7 +67,7 @@ Reconstruction : {
   @table::Reconstruction
   # full MC sequence; form the Calo and CRV matching, and compress based on reco content
   # temporary build surface steps from StepPointMCs; this is a patch
-  MCReco :  [@sequence::CaloMC.TruthMatch,  CrvCoincidenceClusterMatchMC, SelectReco, compressRecoMCs, CrvCoincidenceClusterMCAssns ]
+  MCReco :  [@sequence::CaloMC.TruthMatch, CrvCoincidenceClusterMatchMC, compressRecoMCs, CrvCoincidenceClusterMCAssns ]
 
   # MC truth matched to reco
   HighRecoMCProducts : [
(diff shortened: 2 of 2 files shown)


oksuzian:
# Review of PR #535: Remove duplicate SelectReco calls

## Summary
The PR removes the now-unused `Reconstruction.OffSpillPath` definition and eliminates the redundant `SelectReco` calls that came from inlining `OffSpillPath` into `CentralHelixPath`/`LoopHelixPath`. In `recoMC/prolog.fcl`, it also removes `SelectReco` from `MCReco` (since each base reco path already runs it) and drops the now-orphan `OffSpillPath` override.

## Correctness check

✅ **No external references to `Reconstruction.OffSpillPath`** — searched the whole repo; only usages were the ones this PR removes. `JobConfig/recoMC/OffSpill.fcl` only uses `CentralHelixPath`/`LoopHelixPath`, so it's unaffected.

✅ **Dropping `SelectReco` from `MCReco` is safe** — `MCReco` is only ever appended onto a base reco path (`NoFieldPath`, `OnSpillPath`, `CentralHelixPath`, `LoopHelixPath`), and every one of those paths already ends with `SelectReco`. So `SelectReco` still runs before any MC truth-matching module that needs its outputs (e.g. the `strawDigiMCIndexMapTag : "SelectReco:..."` consumer, `compressRecoMCs`).

✅ **The new `CentralHelixPath`/`LoopHelixPath` now match the style of `OnSpillPath`** (`PreTrk, …Trk, <filter>, SelectReco`), which is a nice consistency win.

## Issues / things to flag

### 1. ⚠️ Subtle behavior change in module ordering (worth confirming with author)
This is more than just a dedup — the *effective* order changes.

Before, `Reconstruction.CentralHelixPath` expanded to:
```
PreTrk, OffSpillTrk, SelectReco, CHFilter, SelectReco
```
Since art only runs a module once per path, the effective order was: **`SelectReco` before `CHFilter`**.

After the PR:
```
PreTrk, OffSpillTrk, CHFilter, SelectReco
```
Effective order: **`CHFilter` before `SelectReco`**. Same for `LoopHelixPath`/`LHFilter`.

If `CHFilter` or `LHFilter` were consuming `SelectReco` output (e.g. filtering on selected KalSeeds), they'll now see the un-selected upstream collections instead. This matches what `OnSpillPath` always did (`LHFilter, SelectReco`), so it's likely the intended/correct order, but it is a semantic change and not purely cosmetic as the PR description implies. Worth the author confirming that `CHFilter`/`LHFilter` operate on pre-SelectReco collections.

### 2. Nit: PR description
The description says "removes unused definition (OffSpillPath) and redundant calls to SelectReco." Since the ordering change above is a real behavioral tweak (not just removing a duplicate art would skip), it would be worth mentioning that CHFilter/LHFilter now run before SelectReco rather than after.

### 3. Nit: blank line removed
In `recoMC/prolog.fcl` the removal of the `OffSpillPath` override block leaves the overrides a little more tightly packed — purely cosmetic. No action needed.

## Verdict
LGTM pending confirmation that the `CHFilter`/`LHFilter`-before-`SelectReco` ordering is what's wanted (it is consistent with `OnSpillPath`, so almost certainly yes). No other references to `OffSpillPath` exist in the repo, and `MCReco` still gets `SelectReco` via the base path it's appended to.

```fcl name=JobConfig/reco/prolog.fcl url=https://github.com/Mu2e/Production/blob/aff9d9fe7bb2a55b1a4576636293624c515a684f/JobConfig/reco/prolog.fcl#L365-L380
Reconstruction.OnSpillPath : [
  ...
  SelectReco
]

Reconstruction.CentralHelixPath : [
  @sequence::Reconstruction.PreTrk,
  @sequence::Reconstruction.OffSpillTrk,
  CHFilter,
  SelectReco
]

Reconstruction.LoopHelixPath : [
  @sequence::Reconstruction.PreTrk,
  @sequence::Reconstruction.OffSpillTrk,
  LHFilter,
  SelectReco
]
```

PR #540, opened 2026-05-16
Match the FlateMinus and FlatePlus energy ranges

michaelmackenzie:
The range isn't defined in fcl currently, where the recent MDC2025 dataset appears to use 90 - 110 MeV. This isn't ideal for mu- --> e+ work where we use this sample to model internal RMC spectra. This update matches it to the flat electron configuration, though exact matching between them isn't necessary.


brownd1978 (approved):
I'm OK with this change, but I believe the final values get set in production submission, so check with Yuri or Sophie.

sophiemiddleton:
Yes, they can be overridden in production, thats true

PR #553, opened 2026-07-08
update conditions set for new CRV geometry

rlcee:
This changes the database conditions set from Sim_best v1_4 to v1_5.  I believe this change is only to accommodate a change to the CRV extracted geometry.  I see it is included in EventNtuple, Offline/CrvResponse/test, and mu2e-trig-config/test so I added a few reviewers.  Maybe @bonventre can advise how to migrate those uses away from this validation fcl


AndrewEdmonds11 (approved):
Confirmed with Richie that it should be backwards compatible

PR #555, opened 2026-07-10
Neutron tail simulation

michaelmackenzie:
Add neutron tail primaries for calo cluster studies, add missing product line. This is for calo cluster background studies and is connected to Offline PR https://github.com/Mu2e/Offline/pull/1878


brownd1978 (approved):
This looks fine to me, but someone more familiar with the production use case should review it too.

PR #561, opened 2026-07-27
Follow changes in Offline PR 1900

brownd1978:
This updates regrowing configuration for changes introduced in PR1900.

### DIFF
--- a/JobConfig/cosmic/S2Resampler.fcl
+++ b/JobConfig/cosmic/S2Resampler.fcl
@@ -31,6 +31,9 @@ outputs : { PrimaryOutput :@local::Primary.PrimaryOutput }
 physics.producers.g4run.inputs: @local::Cosmic.S2Inputs
 # Kill cosmics outside of the CRV region
 physics.producers.g4run.Mu2eG4CommonCut: @local::Cosmic.Mu2eG4CommonCutCosmicS2
+# lower the threshold for saving MC Trajectories to cover reflections
+physics.producers.g4run.TrajectoryControl.mcTrajectoryMomentumCut : 40 # MeV/c
+physics.producers.g4run.TrajectoryControl.saveTrajectoryMomentumCut : 40 # MeV/c
 # define sensitive volumes
 physics.producers.g4run.SDConfig.preSimulatedHits:  ["CosmicResampler:CRV"]
 # setup primary finding for either generator (needs extension if new cosmic generators are added)
--- a/JobConfig/cosmic/S2Resampler.fcl
+++ b/JobConfig/cosmic/S2Resampler.fcl
@@ -31,9 +31,6 @@ outputs : { PrimaryOutput :@local::Primary.PrimaryOutput }
 physics.producers.g4run.inputs: @local::Cosmic.S2Inputs
 # Kill cosmics outside of the CRV region
 physics.producers.g4run.Mu2eG4CommonCut: @local::Cosmic.Mu2eG4CommonCutCosmicS2
-# lower the threshold for saving MC Trajectories to cover reflections
-physics.producers.g4run.TrajectoryControl.mcTrajectoryMomentumCut : 40 # MeV/c
-physics.producers.g4run.TrajectoryControl.saveTrajectoryMomentumCut : 40 # MeV/c
 # define sensitive volumes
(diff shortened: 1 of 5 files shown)


oksuzian:
**1. [S1] The new S2Resampler trajectory cuts are silently clobbered by `primary/epilog.fcl` — the change is a no-op.**

  `S2Resampler.fcl` sets the overrides at lines 35-36, but then includes `Production/JobConfig/primary/epilog.fcl` at line 58, whose line 6 is:
  physics.producers.g4run.TrajectoryControl: @local::mu2eg4DefaultTrajectories
  That is a **full-table assignment**, and in FHiCL a later table assignment replaces the entire table — so both `mcTrajectoryMomentumCut` and `saveTrajectoryMomentumCut` are reset to the Offline default of 50 MeV/c. The job runs fine and CI stays green (nothing in the
  validation suite runs S2Resampler), but the stated purpose — saving reflection trajectories down to 40 MeV/c — is not achieved.

  *Fix:* move the two override lines below the epilog includes (after line 59). The neighboring `g4run.inputs` / `Mu2eG4CommonCut` overrides survive only because no epilog happens to touch those keys — worth a one-line comment so the next person doesn't trip on the same
  layering.

  **2. [S2 — pre-existing, one character, please fix here] `outputs.RegrowOutout.fileName` typo defeats the MC output name.**

  `recoMC/regrowLH.fcl:7` (context line right below this PR's edits) overrides `RegrowOutout` — the real module is `RegrowOutput`. The override lands on a stray table nobody references, so the MC regrow job writes `rec.owner.RegrowLH.version.sequencer.art` (the data-tier name
  from the base fcl) instead of `mcs....` — wrong data tier for SAM registration. The whole point of the recoMC wrapper is (a) the Assns override — fixed by this PR — and (b) the mcs filename — still broken. Same typo in `recoMC/regrowKL.fcl:6`.

  **3. [S2] The KL twin was left with the exact bug this PR fixes for LH.**

  `recoMC/regrowKL.fcl:5` still says `Reconstruction.producers.RegrowKL.KalSeedMCAssns : "SelectReco"` — the dead prolog-table override path — so MC regrowKL still never configures its Assns. Since this PR exists to sweep up regrow config, fixing the twin
  (`physics.producers.RegrowKL.KalSeedMCAssns : "SelectReco"` — note `RegrowKinematicLine` is still an `EDProducer` at Offline main, so it stays under `producers`) belongs here. For the record, `reco/regrowKL.fcl` itself still validates against Offline main —
  `RegrowKinematicLine` gained no new mandatory parameters, and `Mu2eKinKal.REGROW` / `KLDriftFit.KKFitSettings` / `LHDRIFTXTRAP` all still exist.

  **4. [S3 — confirm intent] Removing `drop *_*_*_Digitize` is asymmetric with mixed inputs, and the surrounding keep-list is largely decorative.**

  - Mixed digi files carry process name `Mix`, and `"drop *_*_*_Mix"` remains — so for a mixed input the regrow output still drops Digitize-era products (only EWM/PBI/IntensityInfo are re-kept), while an unmixed input now keeps them all (ProtonBunchTime, Digitize
  TriggerResults, trigger-merger products, ...). If regrowing mixed mcs files is in scope, the two input types now produce differently-stocked outputs — confirm that's intended.
  - Pre-existing but adjacent: `Reconstruction.RegrowOutput.outputCommands` has **no leading `"drop *_*_*_*"`** (unlike `Reconstruction.Output`), and art's default disposition is *keep*. So the keep sequences are mostly decorative — everything not from `Mix`/`Reconstruct` is
  kept by default, including the transient `RegrowSH` ComboHits from the regrow process itself. If the list is meant to be curated (its shape strongly suggests a global drop was assumed), add the leading drop; if default-keep is intended, the keeps could be pruned.

brownd1978:
> **1. [S1] The new S2Resampler trajectory cuts are silently clobbered by `primary/epilog.fcl` — the change is a no-op.**
fixed
> **2. [S2 — pre-existing, one character, please fix here] `outputs.RegrowOutout.fileName` typo defeats the MC output name.**
fixed
> **3. [S2] The KL twin was left with the exact bug this PR fixes for LH.**
fixed
> **4. [S3 — confirm intent] Removing `drop *_*_*_Digitize` is asymmetric with mixed inputs, and the surrounding keep-list is largely decorative.**
fixed

PR #562, opened 2026-08-03
Added IPAEndRingStops to track stopped muons in the MuBeamResampler.fcl

Etho-b02:
- Dependent on Offline/Mu2eG4/geom/protonAbsorber_cylindrical_v05.txt
- geom_run1_a.txt shuold move to v05 to reflect correct IPAEndRing dimensions
- Currently EndRingPath is disabled from trigger and out paths in the fcl file
	- To enable paths add IPAEndRingStopPath to the end of physics.trigger_paths and IPAEndRingStopOutput to physics.outPath

### DIFF
--- a/JobConfig/pileup/prolog.fcl
+++ b/JobConfig/pileup/prolog.fcl
@@ -83,6 +83,16 @@ Pileup: {
       verbosityLevel: 1
     }
 
+    IPAEndRingMuonFinder : {
+      module_type : StoppedParticlesFinder
+      particleInput : "g4run"
+      useEventLevelVolumeInfo: true
+      physVolInfoInput: "g4run:eventlevel"
+      stoppingMaterial : "IPAPolystyrene"
+      particleTypes : [ 13, -13 ]
+      verbosityLevel: 1
+    }
+
     stoppedMuonDaughters: {
       module_type: SimParticleDaughterSelector
       particleInput: "TargetMuonFinder"
@@ -122,12 +132,18 @@ Pileup: {
       particleInputs : [ "TargetPiStopFilter" ]
     }
 
-
     compressPVIPAStops: {
       module_type: CompressPhysicalVolumes
       volumesInput : "g4run"
       hitInputs : []
       particleInputs : [ "IPAStopFilter" ]
+    }
+
+    compressPVIPAEndRingStops: {
+      module_type: CompressPhysicalVolumes
+      volumesInput : "g4run"
+      hitInputs : []
+      particleInputs : [ "IPAEndRingStopFilter" ]
     }
 
     protonTimeOffset : @local::CommonMC.protonTimeOffset
@@ -177,12 +193,18 @@ Pileup: {
       mainSPPtrInputs: [ "TargetPionFinder" ]
     }
 
-
     IPAStopFilter: {
       module_type: FilterG4Out
       mainHitInputs: []
       extraHitInputs: [ "g4run:virtualdetector" ]
       mainSPPtrInputs: [ "IPAMuonFinder" ]
+    }
+
+    IPAEndRingStopFilter: {
+      module_type: FilterG4Out
+      mainHitInputs: []
+      extraHitInputs: [ "g4run:virtualdetector" ]
+      mainSPPtrInputs: [ "IPAEndRingMuonFinder" ]
     }
 
     FlashFilter: {
@@ -210,7 +232,8 @@ Pileup: {
       wrapFiles: true
       mu2e: {
         writeEventIDs : true
-        MaxEventsToSkip: @nil
+        //MaxEventsToSkip: @nil
+        MaxEventsToSkip: 0
         debugLevel : 0
         products: {
           genParticleMixer: { mixingMap: [ [ "generate", "" ] ] }
@@ -235,7 +258,8 @@ Pileup: {
       wrapFiles: true
       mu2e: {
         writeEventIDs : true
-        MaxEventsToSkip:  @nil
+        //MaxEventsToSkip:  @nil
+        MaxEventsToSkip: 0
         debugLevel : 0
         products: {
           genParticleMixer: { mixingMap: [ [ "generate", "" ] ] }
--- a/JobConfig/pileup/prolog.fcl
+++ b/JobConfig/pileup/prolog.fcl
@@ -232,7 +232,7 @@ Pileup: {
       wrapFiles: true
       mu2e: {
         writeEventIDs : true
-        MaxEventsToSkip: @nill
+        MaxEventsToSkip: @nil
         debugLevel : 0
         products: {
           genParticleMixer: { mixingMap: [ [ "generate", "" ] ] }
--- a/JobConfig/pileup/prolog.fcl
+++ b/JobConfig/pileup/prolog.fcl
@@ -232,7 +232,6 @@ Pileup: {
       wrapFiles: true
       mu2e: {
         writeEventIDs : true
-        //MaxEventsToSkip: @nil
         MaxEventsToSkip: 0
         debugLevel : 0
         products: {
@@ -258,8 +257,7 @@ Pileup: {
       wrapFiles: true
       mu2e: {
         writeEventIDs : true
-        //MaxEventsToSkip:  @nil
-        MaxEventsToSkip: 0
+        MaxEventsToSkip:  @nil
         debugLevel : 0
         products: {
           genParticleMixer: { mixingMap: [ [ "generate", "" ] ] }
--- a/JobConfig/pileup/prolog.fcl
+++ b/JobConfig/pileup/prolog.fcl
@@ -232,7 +232,7 @@ Pileup: {
       wrapFiles: true
       mu2e: {
         writeEventIDs : true
-        MaxEventsToSkip: 0
+        MaxEventsToSkip: @nill
         debugLevel : 0
         products: {
           genParticleMixer: { mixingMap: [ [ "generate", "" ] ] }
--- a/JobConfig/pileup/MuBeamResampler.fcl
+++ b/JobConfig/pileup/MuBeamResampler.fcl
@@ -38,8 +38,8 @@ physics: {
   IPAEndRingStopPath : [ @sequence::Pileup.beamResamplerSequence, @sequence::Common.g4Sequence, IPAEndRingMuonFinder, IPAEndRingStopFilter, compressPVIPAEndRingStops]
   flashPath : [ @sequence::Pileup.beamResamplerSequence, @sequence::Common.g4Sequence, @sequence::Pileup.flashSequence, @sequence::Pileup.DetStepSequence ]
   earlyFlashPath : [ @sequence::Pileup.beamResamplerSequence, @sequence::Common.g4Sequence, @sequence::Pileup.flashSequence, @sequence::Pileup.EarlyDetStepSequence ]
-  trigger_paths: [ flashPath, earlyFlashPath, targetStopPath, polyStopPath, IPAStopPath, IPAEndRingStopPath ]
-  outPath : [ FlashOutput, EarlyFlashOutput, TargetStopOutput, PolyStopOutput, IPAStopOutput, IPAEndRingStopOutput ]
+  trigger_paths: [ flashPath, earlyFlashPath, targetStopPath, polyStopPath, IPAStopPath ]
+  outPath : [ FlashOutput, EarlyFlashOutput, TargetStopOutput, PolyStopOutput, IPAStopOutput ]
   end_paths: [outPath]
 }
 
@@ -72,12 +72,12 @@ outputs: {
   }
 
   IPAEndRingStopOutput : {
-    module_type : RootOutput
+    module_type: RootOutput
     outputCommands: [ "drop *_*_*_*",
       @sequence::Pileup.SimKeptProducts
      ]
-     SelectEvents : [IPAEndRingStopPath]
-     fileName : "sim.owner.IPAEndRingStops.version.sequencer.art"
+     SelectEvents: [IPAEndRingStopPath]
+     fileName: "sim.owner.IPAEndRingStops.version.sequencer.art"
   }
 
   FlashOutput : {
--- a/JobConfig/pileup/MuBeamResampler.fcl
+++ b/JobConfig/pileup/MuBeamResampler.fcl
@@ -35,10 +35,11 @@ physics: {
   targetStopPath : [ @sequence::Pileup.beamResamplerSequence, @sequence::Common.g4Sequence, TargetStopPrescaleFilter, TargetMuonFinder, TargetStopFilter, compressPVTargetStops]
   polyStopPath : [ @sequence::Pileup.beamResamplerSequence, @sequence::Common.g4Sequence, PolyStopPrescaleFilter, PolyMuonFinder, PolyStopFilter, compressPVPolyStops]
   IPAStopPath : [ @sequence::Pileup.beamResamplerSequence, @sequence::Common.g4Sequence, IPAMuonFinder, IPAStopFilter, compressPVIPAStops]
+  IPAEndRingStopPath : [ @sequence::Pileup.beamResamplerSequence, @sequence::Common.g4Sequence, IPAEndRingMuonFinder, IPAEndRingStopFilter, compressPVIPAEndRingStops]
   flashPath : [ @sequence::Pileup.beamResamplerSequence, @sequence::Common.g4Sequence, @sequence::Pileup.flashSequence, @sequence::Pileup.DetStepSequence ]
   earlyFlashPath : [ @sequence::Pileup.beamResamplerSequence, @sequence::Common.g4Sequence, @sequence::Pileup.flashSequence, @sequence::Pileup.EarlyDetStepSequence ]
-  trigger_paths: [ flashPath, earlyFlashPath, targetStopPath, polyStopPath, IPAStopPath ]
-  outPath : [ FlashOutput, EarlyFlashOutput, TargetStopOutput, PolyStopOutput, IPAStopOutput ]
+  trigger_paths: [ flashPath, earlyFlashPath, targetStopPath, polyStopPath, IPAStopPath, IPAEndRingStopPath ]
+  outPath : [ FlashOutput, EarlyFlashOutput, TargetStopOutput, PolyStopOutput, IPAStopOutput, IPAEndRingStopOutput ]
   end_paths: [outPath]
 }
 
@@ -68,6 +69,15 @@ outputs: {
     ]
     SelectEvents: [IPAStopPath]
     fileName : "sim.owner.IPAStops.version.sequencer.art"
+  }
+
+  IPAEndRingStopOutput : {
+    module_type : RootOutput
+    outputCommands: [ "drop *_*_*_*",
+      @sequence::Pileup.SimKeptProducts
+     ]
+     SelectEvents : [IPAEndRingStopPath]
+     fileName : "sim.owner.IPAEndRingStops.version.sequencer.art"
   }
 
   FlashOutput : {
--- a/JobConfig/pileup/MuBeamResampler.fcl
+++ b/JobConfig/pileup/MuBeamResampler.fcl
@@ -75,9 +75,9 @@ outputs: {
     module_type: RootOutput
     outputCommands: [ "drop *_*_*_*",
       @sequence::Pileup.SimKeptProducts
-     ]
-     SelectEvents: [IPAEndRingStopPath]
-     fileName: "sim.owner.IPAEndRingStops.version.sequencer.art"
+    ]
+    SelectEvents: [IPAEndRingStopPath]
+    fileName: "sim.owner.IPAEndRingStops.version.sequencer.art"
   }
 
   FlashOutput : {


JobConfig/pileup/prolog.fcl
```diff
@@ -235,7 +258,8 @@ Pileup: {
       wrapFiles: true
       mu2e: {
         writeEventIDs : true
-        MaxEventsToSkip:  @nil
+        //MaxEventsToSkip:  @nil
+        MaxEventsToSkip: 0
```
AndrewEdmonds11, line 262: I would change this back to ```@nil``` since we just hard-coded it to 0 for our tests

AndrewEdmonds11 (approved):
Looks good to me. I'll leave it up to the Production team to decide whether the extra IPAEndRingStops output wants to be on by default

oksuzian (changes requested):
### PR Review Summary — Mu2e/Production#562 "Add IPAEndRingStops to track stopped muons in MuBeamResample.fcl"

Reviewed at head `f57078ab`

**Decision**
- 🔴 request changes

**Scope understood**
- `JobConfig/pileup/prolog.fcl`: adds `IPAEndRingMuonFinder` (`StoppedParticlesFinder`, `stoppingMaterial: "IPAPolystyrene"`), `IPAEndRingStopFilter` (`FilterG4Out`) and `compressPVIPAEndRingStops`, mirroring the existing IPA chain.
- `JobConfig/pileup/MuBeamResampler.fcl`: adds `IPAEndRingStopPath` to `trigger_paths`, `IPAEndRingStopOutput` (a 6th `RootOutput`) to `outPath`.
- Also in `JobConfig/pileup/prolog.fcl` and **not** described by the PR title/body: `beamResampler` and `neutralsResampler` have `MaxEventsToSkip: @nil` replaced by `MaxEventsToSkip: 0`, with the `@nil` lines left commented out.

**Carry-forward from the existing review**
- @AndrewEdmonds11 (APPROVED, 2026-08-03) left an inline request on `JobConfig/pileup/prolog.fcl:262`: *"I would change this back to `@nil` since we just hard-coded it to 0 for our tests."* → **UNADDRESSED** at `f57078ab` (the PR still has a single commit, authored before the comment). Carried forward as finding 1, and it applies to **both** occurrences, not just line 262.
- @AndrewEdmonds11's approval note — *"I'll leave it up to the Production team to decide whether the extra IPAEndRingStops output wants to be on by default"* — is still an open question. Carried forward as finding 3, with evidence on why the default matters.

**Findings**

1. 🔴 [S0] Leftover test hack: `MaxEventsToSkip: @nil` → `0` in the shared pileup prolog (carried over, unaddressed)
   - Evidence: `JobConfig/pileup/prolog.fcl:235-236` (`beamResampler`) and `:261-262` (`neutralsResampler`) at `f57078ab`:
     ```
     //MaxEventsToSkip: @nil
     MaxEventsToSkip: 0
     ```
     `Pileup.filters.beamResampler` / `.neutralsResampler` are consumed by `MuBeamResampler.fcl`, `EleBeamResampler.fcl`, `PiBeamResampler.fcl` and `NeutralsResampler.fcl`, so this is not scoped to the IPA end-ring work at all.
   - Impact: `@nil` is the Mu2e "you must supply this" sentinel — note `fileNames: @nil` two lines above is deliberately left as `@nil` in the same block. `Offline/EventMixing/src/ResamplingMixer_module.cc` declares `fhicl::Atom<unsigned> maxEventsToSkip{..., 0u}` and `eventsToSkip()` draws `std::uniform_int_distribution<size_t> uniform(0, maxEventsToSkip_)` **once**, to pick the starting offset into the first secondary file. Replacing the sentinel with a legal value `0` converts a hard, loud failure ("you forgot the offset") into a silent one: every job that does not carry an explicit override starts at event 0 of its resampling file, i.e. all jobs resample the *same* events. No released dataset is affected today — the prodtools job-definition path appends `physics.filters.<resampler>.mu2e.MaxEventsToSkip: <N>` after the entry's `fcl_overrides`, and `Validation/nightly/pileup_09.fcl` sets it explicitly (`8934`) — which is precisely why the nightly would not catch a regression here either. The value of the `@nil` is the guard, and this removes it.
   - Suggested fix: restore `MaxEventsToSkip: @nil` in both blocks and delete the commented-out lines rather than leaving them in (no commented-out code). If a hard-coded 0 is wanted for local testing, put it in the test's own top-level fcl (as `Scripts/ListPileupProducts.sh` and `Validation/nightly/pileup_0*.fcl` already do), not in the shared prolog.

2. 🟠 [S1] `stoppingMaterial: "IPAPolystyrene"` matches nothing in any released geometry, and nothing in the geometries production pins — the new stream will select zero events
   - Evidence: `StoppedParticlesFinder::materialAccepted()` (`Offline/CommonMC/src/StoppedParticlesFinder_module.cc`) is an exact string compare `material == stoppingMaterial_`; a non-matching name is silently zero-acceptance, no exception. Tracing the name:
     - `IPAPolystyrene` is defined in `Offline/Mu2eG4/src/ConstructMaterials.cc:1337` and consumed by exactly one geometry file, `Offline/Mu2eG4/geom/protonAbsorber_cylindrical_v05.txt` (`protonabsorber.ipa.endRingMaterial = "IPAPolystyrene"`). Both are on Offline **main** only.
     - `protonAbsorber_cylindrical_v05.txt` is included only by `geom_run1.txt` and `geom_run1_a_stickman.txt` (and therefore `geom_common.txt`) — on main.
     - Every currently released SimJob backing release lacks it entirely: `protonAbsorber_cylindrical_v05.txt` is absent from `/cvmfs/mu2e.opensciencegrid.org/Musings/Offline/{v13_30_00,v13_31_00,v13_32_10,v13_33_00}` (v13_33_00 backs the newest musing, SimJob/MDC2025au). In those releases `geom_common.txt → geom_run1_a_stickman.txt → protonAbsorber_cylindrical_v04.txt`, i.e. `protonabsorber.ipa.endRingMaterial = "CorrugatedPolypropylene"`.
     - Even once main is released, the production `MuBeamResampler` entries override `services.GeometryService.inputFile` to `Offline/Mu2eG4/geom/geom_run1_a.txt` (MDC2025 / MDC2030 campaigns) or `geom_run1_b_v01.txt` / `geom_run1_b_v40.txt` (Run1B), and all of those resolve to `protonAbsorber_cylindrical_v04.txt` at line 66 — `CorrugatedPolypropylene`, not `IPAPolystyrene`.
   - Impact: `IPAEndRingMuonFinder` produces an empty `SimParticlePtrCollection` every event, `IPAEndRingStopFilter` rejects, `IPAEndRingStopPath` never passes, and `IPAEndRingStopOutput` writes an events-free `sim.*.IPAEndRingStops.*.art` for every job. The failure is completely silent — no exception, no non-zero exit — so it would only surface as an empty dataset after a full campaign.
   - Suggested fix: either (a) use the material actually configured for the end rings in the geometry this is meant to run against — `CorrugatedPolypropylene` for the `_v04`-based `geom_run1_*` family — or (b) state in the PR body which Offline version + geometry file this requires (`protonAbsorber_cylindrical_v05.txt`, i.e. `geom_common.txt`/`geom_run1.txt` on main) and confirm the production entries will be repointed. Please also post the `StoppedParticlesFinder` end-of-job line (`accepted = N`) from a run against the intended geometry as validation evidence — a non-zero `accepted` is the one-line proof this finding is closed.

3. 🟡 [S2] The 6th output stream is unconditional in a shared production fcl (@AndrewEdmonds11's open question)
   - Evidence: `MuBeamResampler.fcl:41-42` puts `IPAEndRingStopPath` in `trigger_paths` and `IPAEndRingStopOutput` in `outPath` unconditionally; `MuBeamResampler.fcl:74-81` adds the `RootOutput`. This fcl is the production beam-flash workhorse (MDC2025, MDC2030, Run1B all drive it), and each job now writes one more art file.
   - Impact: every future `MuBeamFlash` job in every campaign gains a 6th output that the production tooling will register in SAM. With finding 2 unresolved, those are O(10^4) events-free files per campaign. There is also no downstream consumer yet — nothing cats or selects `IPAEndRingStops` the way `IPAStopsCat` / `Scripts/ListPileupProducts.sh` handle `IPAStops` — so the dataset is write-only until a follow-up lands.
   - Suggested fix: either confirm with the Production team that the stream should be on by default and add the matching downstream entry, or keep `IPAEndRingStopPath` / `IPAEndRingStopOutput` out of the default `trigger_paths` / `outPath` and let a campaign entry opt in via `fcl_overrides`. `JobConfig/pileup/IPAStopSelector.fcl` is the precedent for a separately-driven IPA stream.

4. ⚪ [S3] Formatting of the new output block
   - Evidence: `MuBeamResampler.fcl:74-81` — `module_type :` / `SelectEvents :` / `fileName :` use a space before the colon and the last three lines are indented 5 spaces, where the four sibling output blocks in the same file use `module_type:` and consistent 4-space bodies.
   - Impact: none functionally; it just reads as a different hand in an otherwise uniform file.
   - Suggested fix: match the surrounding `IPAStopOutput` block verbatim.

**Verified — no action needed**
- 🟢 Producer/filter/compressor wiring is internally consistent and mirrors the IPA chain exactly: `IPAEndRingMuonFinder` → `IPAEndRingStopFilter.mainSPPtrInputs` → `compressPVIPAEndRingStops.particleInputs: ["IPAEndRingStopFilter"]` → `IPAEndRingStopOutput.SelectEvents: [IPAEndRingStopPath]`. Every label referenced by the new path is defined in the new prolog blocks.
- 🟢 No duplicated simulation cost from the new path: `IPAEndRingStopPath` shares `@sequence::Pileup.beamResamplerSequence` and `@sequence::Common.g4Sequence` with the five existing paths, and art executes each module once per event regardless of how many paths name it.
- 🟢 `@sequence::Pileup.SimKeptProducts` used by the new output exists (`JobConfig/pileup/prolog.fcl:356`) and is the same sequence the sibling stop outputs use.
- 🟢 Build systems: checked both. `Mu2e/Production` has no `CMakeLists.txt` anywhere in the tree and exactly one `SConscript` (`JobConfig/ensemble/SConscript`, unrelated to fcl); pileup fcl is resolved at run time via `FHICL_FILE_PATH`, so there is no build file to mirror for this change and no `configure_file`/install list to update.
- 🟢 Whitespace-only deletions of the two blank lines at `prolog.fcl` (before `compressPVIPAStops` and before `IPAStopFilter`) are cosmetic and harmless.

**Validation check**
- Build/tests run: none by me; no validation evidence is given in the PR body either. FNALbuild lists `build` as the required test for `JobConfig` changes.
- Config contract check: **partial**. FHiCL structure and module/label wiring are coherent (finding "Verified" above); the `stoppingMaterial` value does not resolve against any released geometry (finding 2); `MaxEventsToSkip` no longer enforces its supply contract (finding 1).
- Cross-repo consistency: **needs follow-up**. The change depends on Offline `protonAbsorber_cylindrical_v05.txt` / `IPAPolystyrene`, which exist on Offline main but in none of the SimJob backing releases up to v13_33_00, and which the production-pinned `geom_run1_*` geometries do not include.

**Residual risk**
- A silently-empty `IPAEndRingStops` dataset produced at campaign scale before anyone notices (finding 2 has no loud failure mode).
- A future resampler entry that omits the `MaxEventsToSkip` override silently resampling identical events across all its jobs (finding 1).
- One extra art file per `MuBeamFlash` job across MDC2025 / MDC2030 / Run1B storage and SAM bookkeeping (finding 3).

**Author follow-ups**
1. Revert `MaxEventsToSkip` to `@nil` in both `beamResampler` and `neutralsResampler`, and delete the commented-out `//MaxEventsToSkip: @nil` lines (finding 1 — already requested inline by @AndrewEdmonds11 and still open).
2. Confirm the intended stopping material: `CorrugatedPolypropylene` (what `protonAbsorber_cylindrical_v04.txt`, and hence every production geometry today, sets for the IPA end rings) vs `IPAPolystyrene` (only in `protonAbsorber_cylindrical_v05.txt` on Offline main). Fix the value or document the Offline/geometry requirement in the PR body (finding 2).
3. Post validation evidence: the `StoppedParticlesFinder` end-of-job summary for `IPAEndRingMuonFinder` (`accepted = ...`) from a short `MuBeamResampler` run against the geometry you intend, plus the resulting file's event count.
4. Decide with the Production team whether the 6th output should be on by default, and if so, say which downstream Cat/selector entry will consume `IPAEndRingStops` (finding 3).
5. Best practice reminder: please add a meaningful PR description covering intent, scope, and validation evidence, and keep the PR to a single topic — the `MaxEventsToSkip` edit is unrelated to the IPA end-ring stops. (Also, the title says `MuBeamResample.fcl`; the file is `MuBeamResampler.fcl`.)

oksuzian:
@FNALbuild run build test

oksuzian (changes requested):
### PR Review Summary — re-review of Mu2e/Production#562 "Add IPAEndRingStops to track stopped muons in MuBeamResample.fcl"

Reviewed at head `ab80665d` (2026-08-10). Second pass. Prior reviews: @AndrewEdmonds11 APPROVED at `f57078ab` with one inline request, and mine 🔴 at `f57078ab` (4887595941, 4 findings). The delta is one commit, `ab80665d` "Adressed PR comments:", touching both files. Every prior finding is verified in the code at this head, not read off the commit message.

**Decision**
- 🔴 Request changes — for one thing only, and it is narrow. Three of the four findings are properly addressed, and the design decision you made on the output stream is the right one. But the `MaxEventsToSkip` fix landed on only one of the two resamplers, and the one still holding the hard-coded `0` is `beamResampler` — the one this PR's own `MuBeamResampler.fcl` actually runs. One line, and I flip to approve.

**Scope understood (unchanged, plus this pass's changes)**
- `IPAEndRingMuonFinder` / `IPAEndRingStopFilter` / `compressPVIPAEndRingStops` are added to the shared pileup prolog, and `IPAEndRingStopPath` / `IPAEndRingStopOutput` to `MuBeamResampler.fcl`.
- New this pass: the path and output are deliberately kept **out** of `trigger_paths` and `outPath`; `neutralsResampler`'s `MaxEventsToSkip` is restored to `@nil`; the commented-out `//MaxEventsToSkip: @nil` lines are deleted from both blocks; the output block's `key :` spacing is normalised.

---

**Findings**

1. 🔴 [S0] `MaxEventsToSkip: 0` is still in `beamResampler` — the fix landed on `neutralsResampler` only (carried over, **PARTIAL**).
   - Evidence: `JobConfig/pileup/prolog.fcl` at `ab80665d`:
     ```
     228:    beamResampler: {
     235:        MaxEventsToSkip: 0          <-- still hard-coded
     253:    neutralsResampler: {
     260:        MaxEventsToSkip:  @nil      <-- restored
     ```
     `main` has `@nil` at both sites (`:213`, `:238`), so the `0` is introduced by this PR and survives at `beamResampler`. The commit message states *"changed MaxEventsToSkip: nil in pileup/prolog : -neutralsResampler -beamResampler"* — the diff (`f57078ab...ab80665d`) shows `beamResampler` only lost its commented-out line; the value was not restored.
   - Impact: this is the more consequential of the two. `Pileup.beamResamplerSequence : [genCounter, protonTimeOffset, beamResampler]` (`prolog.fcl:370`) is what `MuBeamResampler.fcl` runs on every one of its paths, and `EleBeamResampler.fcl` / `PiBeamResampler.fcl` use it too. As in the first review: `@nil` is the "you must supply this" sentinel, and replacing it with the legal value `0` turns a loud startup failure into a silent one — any job without an explicit override starts at event 0 of its resampling file, so all such jobs resample the *same* events. Nothing breaks today (prodtools appends the override, and `Validation/nightly/pileup_09.fcl` sets it explicitly), which is exactly why nothing would catch the regression either.
   - Suggested fix: `MaxEventsToSkip: @nil` at `:235`, matching what you already did at `:260`.
   - Credit where due: @AndrewEdmonds11's inline request pointed at line 262, i.e. the `neutralsResampler` occurrence, and **that one is fixed**. My finding extended it to both blocks; this is the half that remains.

2. 🟡 [S2] `stoppingMaterial: "IPAPolystyrene"` still will not match under any geometry production actually runs — now dormant rather than harmful (carried over, **downgraded from 🟠 S1, PARTIAL**).
   - What changed: with `IPAEndRingStopPath` out of `trigger_paths`, `IPAEndRingMuonFinder` is not scheduled, so the zero-acceptance behaviour cannot produce empty datasets today. The severity drop reflects that, not a change in the underlying facts.
   - What you answered: the commit message states this needs `protonAbsorber_cylindrical_v05.txt` and the latest geometry files. That is correct and it is the option-(b) answer I asked for — but it is in the commit message, not the PR body, and one gap remains open underneath it.
   - Evidence, re-verified at Offline `main` today:
     - `IPAPolystyrene` end rings come only from `protonAbsorber_cylindrical_v05.txt`, included by exactly two files: `geom_run1.txt:66` and `geom_run1_a_stickman.txt` (hence `geom_common.txt`, the default).
     - **`geom_run1_a.txt:66` is still `protonAbsorber_cylindrical_v04.txt`** — `CorrugatedPolypropylene`. And `geom_run1_b_v01.txt:4` simply `#include`s `geom_run1_a.txt`, so Run1B inherits v04 too.
     - Every production entry that drives this fcl overrides the geometry to one of those: `services.GeometryService.inputFile : Offline/Mu2eG4/geom/geom_run1_a.txt` for MDC2025 and MDC2030 `MuBeamFlash`, and `geom_run1_b_v01.txt` for Run1B.
     - No released Offline musing carries v05 at all: absent from `v13_30_00`, `v13_31_00`, `v13_32_10` and `v13_33_00` (the newest), where `geom_run1_a_stickman.txt` still includes v04.
   - Impact: the dependency on unreleased Offline `main` is real but tracked. The part that is *not* addressed by "wait for the Offline release" is that `geom_run1_a.txt` — the file production actually pins — does not pick up v05 even on `main`. Whoever enables this path later inherits a silently zero-acceptance stream, because `StoppedParticlesFinder::materialAccepted()` is an exact string compare with no diagnostic.
   - Suggested fix: say in the **PR body** which Offline version and geometry file this requires, and state explicitly whether `geom_run1_a.txt` is expected to move to v05 or whether the production entries will be repointed at `geom_run1.txt`. When you enable the path, post the `StoppedParticlesFinder` end-of-job `accepted = N` for `IPAEndRingMuonFinder` — a non-zero count is the one-line proof this is closed.

3. ⚪ [S3] Indentation of the new output block is still off by one (carried over, **PARTIAL**).
   - Evidence: `MuBeamResampler.fcl:74-81` — the `key :` → `key:` spacing is fixed, but the closing `]` and the `SelectEvents` / `fileName` lines are still indented 5 spaces where `IPAStopOutput` (`:65-72`) and the other four blocks use 4.

4. ⚪ [S3] Nothing documents how to turn the stream on, and the PR body is still one line.
   - Enabling it means overriding `physics.trigger_paths` and `physics.outPath`, which are sequences — an `fcl_overrides` entry has to restate the whole list, not append to it. A one-line comment above `IPAEndRingStopPath` giving the two override lines would save the next person the reverse-engineering.
   - The PR body still does not mention the Offline/geometry dependency, the validation status, or that the stream ships disabled. The title still says `MuBeamResample.fcl`; the file is `MuBeamResampler.fcl`.

---

**Carry-forward accounting (vs review 4887595941 at `f57078ab`)**

1. 🔴 [was S0] `MaxEventsToSkip` — **PARTIAL**, see finding 1. `neutralsResampler` fixed and both commented-out lines removed; `beamResampler` value not restored.
2. 🟡 [was S1] `stoppingMaterial` — **PARTIAL, downgraded**, see finding 2. Documented in the commit message and defused by disabling the path; the `geom_run1_a.txt` gap is still open.
3. 🟢 [was S2] **The 6th output stream is no longer on by default — FIXED in `ab80665d`, verified, and this is the right call.** `trigger_paths` and `outPath` at `:41-42` are back to the five original entries. This also resolves @AndrewEdmonds11's open question from his approval note ("I'll leave it up to the Production team to decide whether the extra IPAEndRingStops output wants to be on by default") in the conservative direction: the machinery is staged, the decision stays with the Production team, and no campaign gains a file it did not ask for.
4. 🟡 [was S3] Output-block formatting — **PARTIAL**, see finding 3.

---

**Verified 🟢 — no action needed**

- 🟢 **Leaving `IPAEndRingStopPath` and `IPAEndRingStopOutput` defined but unscheduled is safe — art ignores both silently. Tested, not assumed.** I ran a minimal job under Offline `v13_33_00` (art `v3_15_00`) with a path defined in `physics` but absent from `trigger_paths`, and a `RootOutput` defined in `outputs` but absent from `outPath`, including a `SelectEvents` pointing at the unscheduled path. Result: `Art has completed and will exit with status 0`, only the scheduled output's file was created, the unscheduled one was never opened, and no warning or error naming either was emitted. So the staged blocks are genuinely inert — they cost nothing until someone lists them.
  - Worth knowing for finding 4: because art never validates the unscheduled config, a typo in `IPAEndRingStopPath` or in `SelectEvents` will not surface until the day the stream is enabled.
- 🟢 The producer/filter/compressor wiring still mirrors the IPA chain exactly and every label the new path names is defined: `IPAEndRingMuonFinder` → `IPAEndRingStopFilter.mainSPPtrInputs` → `compressPVIPAEndRingStops.particleInputs: ["IPAEndRingStopFilter"]` → `IPAEndRingStopOutput.SelectEvents: [IPAEndRingStopPath]`.
- 🟢 The new prolog blocks land in `Pileup.producers` / `Pileup.filters`, so they are now configured into every fcl including `JobConfig/pileup/prolog.fcl` — `EleBeamResampler`, `NeutralsResampler`, `PiBeamResampler`, `MuStopPileup` and the selectors. Unscheduled everywhere, so harmless; called out because the blast radius of that prolog is wider than this PR's title suggests.
- 🟢 `@sequence::Pileup.SimKeptProducts` used by the new output exists and is the same sequence the sibling stop outputs use.
- 🟢 No build file to update: `Mu2e/Production` has no `CMakeLists.txt` in the tree and one unrelated `SConscript`; pileup fcl resolves at run time via `FHICL_FILE_PATH`.
- 🟢 `neutralsResampler`'s restored `@nil` is spelled exactly as `main` has it, two-space form included.

**Validation check**
- Build/tests run: `mu2e/buildtest` had no result at `ab80665d` ("This test has not been triggered yet"), so I triggered one — this is the required test for `JobConfig` changes, and FNALbuild notes the author is not a Mu2e org member, so it needs a member to start it. **Result pending at `ab80665d` as of this review**; please check it before merging. Separately, I ran the unscheduled-path/output test described above myself.
- Config contract check: **partial** — FHiCL structure and label wiring are coherent and the disabled stream is verified inert; `MaxEventsToSkip` still does not enforce its supply contract at `beamResampler` (finding 1).
- Cross-repo consistency: **needs follow-up** — depends on Offline `main` (`protonAbsorber_cylindrical_v05.txt`), which is in no released musing through `v13_33_00`, and `geom_run1_a.txt` — what production actually pins — does not include it even on `main` (finding 2).

**Residual risk**
- A future resampler entry that omits the `MaxEventsToSkip` override silently resampling identical events across all its jobs (finding 1).
- The IPA end-ring stream is now staged but unvalidated and unvalidatable by art until it is switched on; the first person to enable it inherits both the material mismatch and any typo in the unscheduled config.

**Author follow-ups**
1. `MaxEventsToSkip: @nil` at `prolog.fcl:235` (`beamResampler`) — the one blocking item (finding 1).
2. Put the Offline/geometry dependency in the PR body, and say whether `geom_run1_a.txt` moves to `protonAbsorber_cylindrical_v05.txt` or the production entries get repointed (finding 2).
3. Optional: 4-space indent on the new output block (finding 3), and a comment giving the two overrides needed to enable the stream (finding 4).
4. Check the `mu2e/buildtest` result at `ab80665d` before merging — I triggered it, but it had not reported when this review was written.

oksuzian:
@FNALbuild run build test

oksuzian (approved):

PR #563, opened 2026-08-06
Add script to select and regrow cosmic reflections

brownd1978:
This script shows how to use SelectReflections features introduced in Offline PR 1917

### DIFF
--- a/JobConfig/recoMC/RefRegrow.fcl
+++ b/JobConfig/recoMC/RefRegrow.fcl
@@ -43,6 +43,10 @@ physics.filters : {
 
   RegrowKKe : {
     @table::Mu2eKinKal.RegrowLH
+    ExtrapolationSettings :{
+      @table::Mu2eKinKal.LHDRIFTXTRAP
+      BackToTracker : true
+    }
     KalSeedMCAssns : "SelectReco"
     CopyKalSeedMCs : true
     KalSeedPtrCollection : Reflecte
@@ -52,6 +56,10 @@ physics.filters : {
 
   RegrowKKmu : {
     @table::Mu2eKinKal.RegrowLH
+    ExtrapolationSettings :{
+      @table::Mu2eKinKal.LHDRIFTXTRAP
+      BackToTracker : true
+    }
     KalSeedMCAssns : "SelectReco"
     CopyKalSeedMCs : true
     KalSeedPtrCollection : Reflectmu
--- a/JobConfig/recoMC/RefRegrow.fcl
+++ b/JobConfig/recoMC/RefRegrow.fcl
@@ -1,0 +1,137 @@
+#
+# Script to select and regrow reflecting cosmic ray particles starting with standard recoMC output
+# original author: D. Brown, 6/25/26
+#
+#include "Production/JobConfig/recoMC/prolog.fcl"
+outputs.CentralHelixOutput.fileName : nil
+outputs.LoopHelixOutput.fileName : nil
+services : @local::Services.Reco
+process_name: RefRegrow
+source : { module_type : RootInput }
+
+physics.filters : {
+  eidsel : {
+    module_type : EventIDFilter
+    idsToMatch: [ "*:*:*" ]
+  }
+
+  Reflecte : {
+    @table::TrkReco.filters.SelectT0Reflections
+    UpstreamKalSeedCollection : "KKUe"
+    DownstreamKalSeedCollection : "KKDe"
+  }
+
+  RegrowReflecte : {
+    @table::TrkReco.filters.SelectReflections
+    UpstreamKalSeedCollection : "RegrowKKe"
+    DownstreamKalSeedCollection : "RegrowKKe"
+  }
+
+  Reflectmu : {
+    @table::TrkReco.filters.SelectT0Reflections
+    UpstreamKalSeedCollection : "KKUmu"
+    DownstreamKalSeedCollection : "KKDmu"
+  }
+
+  RegrowReflectmu : {
+    @table::TrkReco.filters.SelectReflections
+    UpstreamKalSeedCollection : "RegrowKKmu"
+    DownstreamKalSeedCollection : "RegrowKKmu"
+  }
+
+  RegrowKKe : {
+    @table::Mu2eKinKal.RegrowLH
+    KalSeedMCAssns : "SelectReco"
+    CopyKalSeedMCs : true
+    KalSeedPtrCollection : Reflecte
+    ComboHitCollection : RegrowSH
+    MustRegrow : "All"
+ }
+
+  RegrowKKmu : {
+    @table::Mu2eKinKal.RegrowLH
+    KalSeedMCAssns : "SelectReco"
+    CopyKalSeedMCs : true
+    KalSeedPtrCollection : Reflectmu
+    ComboHitCollection : RegrowSH
+    MustRegrow : "All"
+  }
+}
+
+physics.producers : {
+  RegrowSH : {
+    @table::TrkHitReco.makeSH
+    StrawDigiCollectionTag  : "SelectReco"
+    StrawDigiADCWaveformCollectionTag : "SelectReco"
+  }
+
+# MC recompression not yet debugged: TODO
+  RefCompressRecoMCse : {
+    @table::Reconstruction.producers.compressRecoMCs
+    strawDigiMCIndexMapTag: "SelectReco:StrawDigiMap"
+    crvDigiMCIndexMapTag: "SelectReco:CrvDigiMap"
+    simParticleTags: [ "compressRecoMCs" ]
+    crvDigiMCTag: "compressRecoMCs"
+    extraStepPointMCTags: [ "compressRecoMCs:virtualdetector" ]
+    mcTrajectoryTag: "compressRecoMCs"
+    primaryParticleTag: "compressRecoMCs"
+    strawDigiMCTag: "compressRecoMCs"
+    surfaceStepTags: [ "compressRecoMCs" ]
+    caloClusterMCTag: "compressRecoMCs"
+    crvCoincClusterMCTags :  [ "compressRecoMCs:CrvCoincidenceClusterMatchMC" ]
+  }
+
+  RefCompressRecoMCsmu : {
+    @table::Reconstruction.producers.compressRecoMCs
+    strawDigiMCIndexMapTag: "SelectReco:StrawDigiMap"
+    crvDigiMCIndexMapTag: "SelectReco:CrvDigiMap"
+    simParticleTags: [ "compressRecoMCs" ]
+    crvDigiMCTag: "compressRecoMCs"
+    extraStepPointMCTags: [ "compressRecoMCs:virtualdetector" ]
+    mcTrajectoryTag: "compressRecoMCs"
+    primaryParticleTag: "compressRecoMCs"
+    strawDigiMCTag: "compressRecoMCs"
+    surfaceStepTags: [ "compressRecoMCs" ]
+    caloClusterMCTag: "compressRecoMCs"
+    crvCoincClusterMCTags :  [ "compressRecoMCs:CrvCoincidenceClusterMatchMC" ]
+  }
+}
+
+physics.RePath : [ eidsel, Reflecte, RegrowSH, RegrowKKe, RegrowReflecte ]
+physics.RmuPath : [ eidsel, Reflectmu, RegrowSH, RegrowKKmu, RegrowReflectmu ]
+
+physics.trigger_paths : [ RePath, RmuPath ]
+
+outputs.RefOutpute : {
+  @table::Reconstruction.Output
+  SelectEvents : [RePath]
+  outputCommands: [ @sequence::Reconstruction.Output.outputCommands,
+  "drop mu2e::*_KK*_*_*",
+  "drop mu2e::KalSeeds_K*_*_*",
+  "drop mu2e::Kal*_SelectReco_*_*",
+  "drop mu2e::PrimaryParticle_*_*_*",
+  "keep mu2e::*_compressRecoMCs_*_*",
+  "keep mu2e::*_RegrowKK*_*_*" ]
+}
+
+outputs.RefOutputmu : {
+  @table::Reconstruction.Output
+  fileName : "/data/HD6/users/brownd/mcs.brownd.RefRegrowmu.art"
+  SelectEvents : [RmuPath]
+  outputCommands: [ @sequence::Reconstruction.Output.outputCommands,
+  "drop mu2e::*_KK*_*_*",
+  "drop mu2e::KalSeeds_K*_*_*",
+  "drop mu2e::Kal*_SelectReco_*_*",
+  "drop mu2e::PrimaryParticle_*_*_*",
+  "keep mu2e::*_compressRecoMCs_*_*",
+  "keep mu2e::*_RegrowKK*_*_*" ]
+}
+
+physics.RefEndPath : [ RefOutpute, RefOutputmu ]
+
+physics.end_paths : [ RefEndPath ]
+
+# Uncomment and update as needed
+#services.DbService.purpose: Sim_best
+#services.DbService.version: v1_4
+#services.DbService.verbose: 2
--- a/JobConfig/recoMC/RefRegrow.fcl
+++ b/JobConfig/recoMC/RefRegrow.fcl
@@ -25,6 +25,7 @@ physics.filters : {
     @table::TrkReco.filters.SelectReflections
     UpstreamKalSeedCollection : "RegrowKKe"
     DownstreamKalSeedCollection : "RegrowKKe"
+    ReflectionTestDirection : "Both" # Require both direction to have a reflection match
   }
 
   Reflectmu : {
@@ -37,6 +38,7 @@ physics.filters : {
     @table::TrkReco.filters.SelectReflections
     UpstreamKalSeedCollection : "RegrowKKmu"
     DownstreamKalSeedCollection : "RegrowKKmu"
+    ReflectionTestDirection : "Both" # Require both direction to have a reflection match
   }
 
   RegrowKKe : {
--- a/JobConfig/recoMC/RefRegrow.fcl
+++ b/JobConfig/recoMC/RefRegrow.fcl
@@ -115,6 +115,7 @@ physics.trigger_paths : [ RePath, RmuPath ]
 outputs.RefOutpute : {
   @table::Reconstruction.Output
   SelectEvents : [RePath]
+  fileName : "mcs.user.RegrownReflectione.config.seq.art"
   outputCommands: [ @sequence::Reconstruction.Output.outputCommands,
   "drop mu2e::*_KK*_*_*",
   "drop mu2e::KalSeeds_K*_*_*",
@@ -126,7 +127,7 @@ outputs.RefOutpute : {
 
 outputs.RefOutputmu : {
   @table::Reconstruction.Output
-  fileName : "/data/HD6/users/brownd/mcs.brownd.RefRegrowmu.art"
+  fileName : "mcs.user.RegrownReflectionmu.config.seq.art"
   SelectEvents : [RmuPath]
   outputCommands: [ @sequence::Reconstruction.Output.outputCommands,
   "drop mu2e::*_KK*_*_*",


oksuzian (changes requested):
### PR Review Summary — Mu2e/Production#563 "Add script to select and regrow cosmic reflections"

Reviewed at head `bec29557`

**Decision**
- 🔴 request changes

**Scope understood**
- Adds one new file, `JobConfig/recoMC/RefRegrow.fcl` (+139/-0), a standalone job that reads standard recoMC output, selects reflecting cosmic-ray track pairs with `SelectT0Reflections`, regrows both members with `RegrowLoopHelix`, and re-tests the regrown pair with the tight tracker-front `SelectReflections`. Separate e and mu paths, separate outputs.
- Exercises the `ReflectionTestDirection` feature added in Offline PR #1917 (still open); the second commit here adds `ReflectionTestDirection : "Both"` to the two regrown-pair filters.
- No prior review or review comment exists on this PR (only FNALbuild status comments), and no prior local draft, so there is nothing to carry forward.

**Findings**

1. 🔴 [S0] `outputs.RefOutpute` never sets `fileName`; the job aborts during module construction
   - Evidence: `JobConfig/recoMC/RefRegrow.fcl:107-117` builds `RefOutpute` from `@table::Reconstruction.Output` and overrides only `SelectEvents` and `outputCommands`. `Reconstruction.Output` resolves through `JobConfig/recoMC/prolog.fcl` onto the base table in `JobConfig/reco/prolog.fcl`, where `fileName : @nil`. Its sibling `RefOutputmu` does set `fileName` (line 121), so the e-output is the odd one out. I reproduced the exact failure with art v3_15_00 / Offline v13_33_00 on a minimal job whose only output inherits `fileName : @nil`:
     ```
     %MSG-s ArtException:  RootOutput:realout@Construction  ModuleConstruction
     ---- Type mismatch BEGIN
       Unsuccessful attempt to convert FHiCL parameter 'fileName' to type
       'std::__cxx11::basic_string<...>'.
       [Specific error:]  can't obtain string from nil
     ---- Type mismatch END
     Art has completed and will exit with status 65.
     ```
   - Impact: the script cannot run at all. It dies before the first event, so neither output is produced and the e-path work is unreachable.
   - Suggested fix: add a placeholder name in the repo convention, e.g. `fileName : "mcs.owner.RefRegrowe.version.sequencer.art"`, matching `JobConfig/recoMC/regrowLH.fcl` and `regrowKL.fcl`.

2. 🟠 [S1] `RefOutputmu.fileName` is a hardcoded personal path
   - Evidence: `RefRegrow.fcl:121` — `fileName : "/data/HD6/users/brownd/mcs.brownd.RefRegrowmu.art"`. Every sibling in this directory uses the placeholder form instead: `JobConfig/recoMC/regrowLH.fcl` has `outputs.RegrowOutput.fileName: "mcs.owner.RegrowLH.version.sequencer.art"`, and `regrowKL.fcl` the same for `RegrowKL`.
   - Impact: that path exists only on the author's machine, so the file as committed fails at output-open for anyone else. It also bypasses the `owner`/`version`/`sequencer` placeholders that `mu2ejobdef` rewrites, so the script cannot be driven from a jobdef for grid running.
   - Suggested fix: `fileName : "mcs.owner.RefRegrowmu.version.sequencer.art"`. (If the intent is instead that a wrapper supplies both names at runtime, then neither output should carry a name — but as written the two disagree with each other, which is what makes this look like an oversight rather than a choice.)

3. 🟠 [S1] Hard dependency on unmerged Offline #1917 — the script cannot parse against Offline `main`
   - Evidence: `RefRegrow.fcl:19` and `:32` reference `@table::TrkReco.filters.SelectT0Reflections`. That table does not exist in `Offline/TrkReco/fcl/prolog.fcl` on `main`: `main` defines only `TrkReco.filters.SelectReflections`, and that one takes `SelectBestPair : 3` as an int and has no `ReflectionTestDirection` key at all. Both `SelectT0Reflections` and the string-valued `SelectBestPair`/`ReflectionTestDirection` appear only at Offline #1917 head `aeb0475`, and that PR is still open.
   - Impact: merging this first lands a script in Production `main` that fails FHiCL resolution on every current Offline. Production's build test does not execute `JobConfig/recoMC/*.fcl`, so nothing in CI would catch it.
   - Suggested fix: note the dependency in the PR body and sequence the merge after Offline #1917.

4. 🟡 [S2] Two dead, mutually identical producer blocks
   - Evidence: `RefRegrow.fcl:71-99` define `RefCompressRecoMCse` and `RefCompressRecoMCsmu` with byte-identical bodies. Neither label appears in `physics.RePath` / `physics.RmuPath` (lines 102-103) nor in `physics.RefEndPath` (line 132), so neither is ever constructed.
   - Impact: ~30 lines of silently inert config. The adjacent `# MC recompression not yet debugged: TODO` comment sets an expectation the file does not meet, and a future reader has to re-derive that these are unused.
   - Suggested fix: drop both until recompression is debugged — git history preserves the draft — or collapse to one block and put it on a path.

5. 🟡 [S2] Vestigial `nil` output lines that do not do what they look like
   - Evidence: `RefRegrow.fcl:6-7` — `outputs.CentralHelixOutput.fileName : nil` and `outputs.LoopHelixOutput.fileName : nil`. This file includes only `Production/JobConfig/recoMC/prolog.fcl`, which defines no `outputs` block, so these two lines *create* two output tables that carry a `fileName` and no `module_type`. They read as "disable the two standard reco outputs", but there is nothing to disable.
   - Impact: none at runtime — I confirmed art v3_15_00 silently ignores an `outputs` entry that is on no end path (test job exited 0 with the stray block present) — but the lines mislead.
   - Suggested fix: delete both lines.

6. 🟡 [S2] `keep mu2e::*_RegrowKK*_*_*` is broader than the established regrow convention (potential risk, please confirm)
   - Evidence: `RefRegrow.fcl:116` and `:129`. `RegrowLoopHelix` registers two products under the same module label — `produces<KKTRKCOL>()` and `produces<KalSeedCollection>()` at `Offline/Mu2eKinKal/src/RegrowLoopHelix_module.cc:179-180`, where `KKTRKCOL` is `OwningPointerCollection<KKTrack<LoopHelix>>`. The reference regrow output in this repo, `Reconstruction.RegrowOutput` in `JobConfig/reco/prolog.fcl`, deliberately keeps only `mu2e::KalSeed*_*_*_Regrow`. No `classes_def.xml` anywhere in Offline declares `KKTrack` or `OwningPointerCollection`.
   - Impact: labeled potential, not observed — I did not confirm the friendly class name that art derives for `KKTRKCOL`, so I cannot say whether the `mu2e::*` class wildcard matches it. If it does, the output is at best far larger than intended and at worst unwritable for want of a dictionary.
   - Suggested fix: narrow both to `keep mu2e::KalSeed*_RegrowKK*_*_*`, which is what the sibling output does and is unambiguously what the job needs.

7. ⚪ [S3] `physics.producers.RegrowSH` duplicates a table the include chain already provides
   - Evidence: `RefRegrow.fcl:64-68` is byte-identical to `Reconstruction.producers.RegrowSH` in `JobConfig/reco/prolog.fcl`, which `recoMC/prolog.fcl` includes.
   - Suggested fix: `RegrowSH : @local::Reconstruction.producers.RegrowSH`.

8. ⚪ [S3] The two output modules carry byte-identical `outputCommands`
   - Evidence: `RefRegrow.fcl:110-116` and `:123-129` differ only in `SelectEvents` and `fileName`.
   - Suggested fix: hoist the command list into one local table and reference it twice, so the two outputs cannot drift apart.

9. ⚪ [S3] `eidsel` is a pass-all filter with no explanation
   - Evidence: `RefRegrow.fcl:12-15` — `EventIDFilter` with `idsToMatch: [ "*:*:*" ]`, first module on both paths. The pattern matches every event, so this is a hook for narrowing to specific event IDs during debugging (same idiom as `Production/Tests/PBI_33344_Pathological.fcl`).
   - Suggested fix: keep it, but add a one-line comment saying it is a debugging hook left wide open, so nobody reads it as a real selection.

- 🟢 Module kinds are right. `SelectReflections` is an `art::EDFilter` (`Offline/TrkReco/src/SelectReflections_module.cc`) and `RegrowLoopHelix` is an `art::EDFilter` (`Offline/Mu2eKinKal/src/RegrowLoopHelix_module.cc`), so putting `Reflecte`/`RegrowReflecte`/`RegrowKKe` and their mu counterparts under `physics.filters` is correct, not a producer-in-filters mistake.
- 🟢 Every overridden key resolves against its intended source. `UpstreamKalSeedCollection`, `DownstreamKalSeedCollection` and `ReflectionTestDirection` all exist in the `SelectReflections` config at #1917 head; `KalSeedPtrCollection`, `ComboHitCollection`, `KalSeedMCAssns`, `CopyKalSeedMCs` and `MustRegrow` all exist in `RegrowLoopHelixConfig`. `MustRegrow : "All"` is one of the three accepted values — the module throws on anything but `None`/`Any`/`All`.
- 🟢 `ReflectionTestDirection : "Both"` is safe here specifically because `TrkReco.filters.SelectReflections` sets `Surface : "TT_Front"`. The constructor throws `"Explicit SID incompatible with t0 segment selection"` whenever no `Surface` is given and the direction is not `Any`, so applying `"Both"` to `SelectT0Reflections` instead would have been fatal. As written, the tight filter is the one that gets it — correct.
- 🟢 Pointing both `UpstreamKalSeedCollection` and `DownstreamKalSeedCollection` at the single `RegrowKKe` collection is coherent, not a copy-paste error: `RegrowKKe` regrows both members of the reflection pair into one `KalSeedCollection`, and the filter separates them by the sign of the t0-segment pz.
- 🟢 `services : @local::Services.Reco` resolves through the single include: `recoMC/prolog.fcl` → `reco/prolog.fcl` → `digitize/prolog.fcl` → `Offline/fcl/standardServices.fcl`, which itself includes `minimalMessageService.fcl`, so the `message : @local::default_message` inside `Services.Core` resolves too.
- 🟢 Both build systems checked. Production has no CMake build at all, and its one `SConscript` (`JobConfig/ensemble/SConscript`) is entirely commented out; `.muse` contributes only `PATH`/`PYTHONPATH` entries. A new `.fcl` under `JobConfig/recoMC/` needs no registration in either system, so there is no un-mirrored build-file change hiding here.

**Validation check**
- Build/tests run: partial, and the green badge is stale. FNALbuild reported all green, but at `5e02871` — the *parent* commit. At head `bec29557`, `mu2e/buildtest` is still `pending` (`jenkins/ghprb` succeeded). In any case Production's build test does not execute `JobConfig/recoMC/RefRegrow.fcl`, so a green run would not have exercised finding 1.
- I ran two targeted art jobs myself (art v3_15_00, Offline v13_33_00) to pin down the two config behaviors asserted above: the `@nil` `fileName` abort, and art's tolerance of an `outputs` entry that is on no end path.
- Config contract check: fail — finding 1 is a hard construction-time abort.
- Cross-repo consistency: needs follow-up — depends on Offline #1917, still open.

**Residual risk**
- The script has almost certainly never been run end to end in its committed form, since it cannot construct its outputs. Whatever the author ran locally differed from what is in the PR, so nothing here is validation evidence for the committed file.
- Merge ordering is unguarded: nothing in Production CI notices that this file needs Offline #1917.
- The MC recompression the file's own TODO calls for is absent, so downstream MC products are the input file's, not recompressed against the regrown tracks. Fine as a staged plan, but worth stating in the PR body.

**Author follow-ups**
1. Add a `fileName` to `RefOutpute` in the `mcs.owner.<desc>.version.sequencer.art` form (finding 1).
2. Replace the hardcoded `/data/HD6/users/brownd/...` path in `RefOutputmu` with the same placeholder form (finding 2).
3. Note the Offline #1917 dependency in the PR body and merge after it (finding 3).
4. Remove the two unused `RefCompressRecoMCs*` blocks and the two `nil` output lines (findings 4, 5).
5. Confirm whether `keep mu2e::*_RegrowKK*_*_*` is meant to pull in the `KKTRKCOL` product, or narrow it to `mu2e::KalSeed*_RegrowKK*_*_*` (finding 6).
6. Best practice reminder: please add validation evidence to the PR description — the exact command, the input dataset, and how many events survived each path — once the output naming is fixed and the script runs.

oksuzian (commented):
### PR Review Summary — second pass on Mu2e/Production#563, "Add script to select and regrow cosmic reflections"

Reviewed at head `987e3b21` (was `bec29557` for the first pass). Delta is two commits: `19b99ade` "Regrow reflections extrapolating back to the tracker" and `987e3b21` "Add output file name". Both blocking findings from the first pass are fixed, and the cross-repo dependency resolved itself when Offline #1917 merged.

**Decision**
- 🟡 Comment only. Upgraded from 🔴.

---

### Carry-forward accounting (vs review 4887610456 at `bec29557`)

**1. 🟢 [was S0] `RefOutpute` missing `fileName` — FIXED in `987e3b21`, verified.**
`RefRegrow.fcl:118` now sets `fileName : "mcs.user.RegrownReflectione.config.seq.art"`. The job no longer aborts at module construction. This was the finding that made the script unrunnable, and it is genuinely resolved rather than worked around.

**2. 🟢 [was S1] Hardcoded personal path in `RefOutputmu` — FIXED in `987e3b21`, verified.**
`RefRegrow.fcl:130` replaces `/data/HD6/users/brownd/mcs.brownd.RefRegrowmu.art` with `mcs.user.RegrownReflectionmu.config.seq.art`. No absolute paths remain in the file.

One correction to my own finding, since I prescribed a specific spelling and you used a different one. I wrote "use the `mcs.owner.<desc>.version.sequencer.art` form", and you used `mcs.user.<desc>.config.seq.art`. **Your version works** — I checked rather than assumed. `Mu2eName` parses dot-names *positionally* (6 fields: tier.owner.description.dsconf.sequencer.extension, `utils/job_common.py:29-54`), so the placeholder words carry no meaning to the tooling; `user`/`config`/`seq` land in the right slots. Worth knowing that prodtools' `jobfcl` has a guard, `_PLACEHOLDER_TOKEN_RE = (description|desc|owner|version|sequencer)` (`utils/jobfcl.py:20`), that raises if an *unsubstituted* placeholder survives into a built job. Your tokens are outside that set, so they will not trip the guard — which cuts both ways: correct substitution passes silently, but a genuinely-unsubstituted name would also pass silently rather than failing loudly. Only relevant if this script is ever driven from a jobdef. As a standalone script it is fine, and the sibling convention (`regrowLH.fcl:6`, `regrowKL.fcl:6`, `OnSpill.fcl:14` all use `owner`/`version`/`sequencer`) is a consistency nit at most, not a defect. Not carrying it as a finding.

**3. 🟢 [was S1] Hard dependency on unmerged Offline #1917 — RESOLVED externally, verified.**
#1917 merged at 2026-08-09T00:42:34Z (merge commit `99536c11`). Re-checked every referenced key against Offline `main`, not against the PR branch:
- `TrkReco.filters.SelectT0Reflections` — present, `TrkReco/fcl/prolog.fcl:149`.
- `TrkReco.filters.SelectReflections` — present, `:159`, with `Surface : "TT_Front"` at `:166`.
- `ReflectionTestDirection` — present in both tables, string-valued.
- `Mu2eKinKal.RegrowLH` — present, `Mu2eKinKal/fcl/prolog.fcl:623`.
- `Mu2eKinKal.LHDRIFTXTRAP` — present, `:374`.

The merge-ordering hazard I raised is gone. Production `main` can now resolve this file.

**4. 🟡 [was S2] Two dead `RefCompressRecoMCs*` blocks — UNADDRESSED, carried over.** `RefRegrow.fcl:78-106`, still byte-identical to each other, still absent from `RePath`/`RmuPath` (`:110-111`) and `RefEndPath` (`:141`). ~30 lines of inert config under a `# MC recompression not yet debugged: TODO`.

**5. 🟡 [was S2] Vestigial `nil` output lines — UNADDRESSED, carried over.** `RefRegrow.fcl:6-7`. These *create* two output tables rather than disabling anything, since the include chain defines no `outputs` block. Harmless at runtime, misleading to read.

**6. 🟡 [was S2] `keep mu2e::*_RegrowKK*_*_*` breadth — UNADDRESSED, and I can now confirm it is real rather than potential.** See finding 1 below; upgrading from "please confirm" to an evidenced finding.

**7. ⚪ [was S3] `RegrowSH` duplicates `Reconstruction.producers.RegrowSH`** — unaddressed, non-gating.

**8. ⚪ [was S3] Byte-identical `outputCommands` in the two outputs** — unaddressed, non-gating.

**9. ⚪ [was S3] `eidsel` pass-all filter needs a comment** — unaddressed, non-gating.

---

### Findings

**1. 🟡 [S2] `keep mu2e::*_RegrowKK*_*_*` writes the full `KKTrack` collection, not just the `KalSeed`s — now confirmed, was "potential" last pass.**

- Evidence: `RegrowLoopHelix` declares **three** products under one label (`Mu2eKinKal/src/RegrowLoopHelix_module.cc:179-189` on merged `main`): `produces<KKTRKCOL>()`, `produces<KalSeedCollection>()`, and conditionally `produces<KalSeedMCCollection>()`. `KKTRKCOL` is `OwningPointerCollection<KKTrack<KTRAJ>>` (`:109-110`). The `mu2e::*` class wildcard matches all three. Ordering makes this reachable: the `keep` is the **last** entry in `outputCommands` (`RefRegrow.fcl:125`, `:137`), and art applies commands in order with last-match-wins, so it overrides the earlier `drop mu2e::*_KK*_*_*`.
- What I could not determine, stated as an assumption: whether `KKTRKCOL` has a ROOT dictionary. I found no `classes_def.xml` entry for `KKTrack` or `OwningPointerCollection` anywhere in Offline. If there is genuinely no dictionary, the output module throws at write time; if there is one, the file is simply much larger than intended, since `SaveTrajectory : Full` (`Mu2eKinKal/fcl/prolog.fcl:637`) makes these objects heavy. I could not run the job to distinguish these, which is why this is S2 and not higher.
- Impact: either an unwritable output or a large one. Every established output in this repo narrows the class: `Reconstruction.Output` uses `keep mu2e::KalSeed*_KKDe_*_*` and `keep mu2e::KalSeed*_KKLine_*_*` (`JobConfig/reco/prolog.fcl:412-413`). This file is the outlier.
- Suggested fix: `keep mu2e::KalSeed*_RegrowKK*_*_*` in both outputs, which matches the sibling convention and is unambiguously what the job needs. If the `KKTrack` collection *is* wanted, say so in the PR body — it is a surprising thing to persist.

**2. ⚪ [S3] `ExtrapolationSettings` override is correct — one confirming note, no action.**

- The new blocks at `:45-48` and `:59-62` re-compose `@table::Mu2eKinKal.LHDRIFTXTRAP` and set `BackToTracker : true`. I verified this is necessary rather than redundant: `RegrowLH` on `main` composes `LHDRIFTXTRAP` bare (`Mu2eKinKal/fcl/prolog.fcl:639-641`), and `LHDRIFTXTRAP` carries `BackToTracker : false` (`:382`). So the override is what supplies the return-leg intersection.
- Recording this because it is the crux of the design and it took me two passes on the Offline side to get right: the tight `TrkReco.filters.SelectReflections` (`MaxDeltaT : 5`, `Surface : "TT_Front"`) is applied only to `RegrowKKe`/`RegrowKKmu` (`:26-27`, `:39-40`) — the *regrown* collections that set the flag locally — while stage 1 uses `SelectT0Reflections`, which has no `Surface` key and matches on t0 alone. The three-stage structure is coherent and the flag is in the right place.

---

### 🟢 Verified — no action needed

- 🟢 Both first-pass blockers fixed and re-verified at head, not taken on the commit messages.
- 🟢 Offline #1917 merged; all five referenced prolog tables resolve against Offline `main`.
- 🟢 `BackToTracker : true` override is necessary, correctly scoped to the regrow stage, and consistent with the merged Offline semantics.
- 🟢 Everything cleared in the first pass still holds: module kinds (`SelectReflections` and `RegrowLoopHelix` are both `EDFilter`s, so `physics.filters` is right), `MustRegrow : "All"` is an accepted value, pointing both Up/Downstream collections at one regrown collection is coherent, `services : @local::Services.Reco` resolves through the single include, and Production needs no build-file registration for a new `.fcl`.

### Validation check
- Build/tests: **`mu2e/buildtest` is `pending` at `987e3b21`** (`jenkins/ghprb` success). The last green was at `5e028717`, two commits back. I did **not** trigger a run — one is already in flight at this head, and re-requesting would only queue a duplicate. Note Production's build test does not execute `JobConfig/recoMC/*.fcl`, so a green result would not exercise any finding here.
- Config contract: **pass** — the construction-time abort is gone, and every `@table::`/`@local::` reference resolves against merged Offline `main`.
- Cross-repo consistency: **pass** — the #1917 dependency is satisfied.
- Physics validation: none run by me, and none in CI. Unchanged from the first pass.

### Residual risk
- The script still has no stated validation evidence. With the output naming fixed it should now run, but nothing in the PR says it has been run end to end in its committed form, and no CI job executes it.
- Finding 1 is unresolved and would surface on first execution — either as a write-time throw or as an unexpectedly large output. That is a fast, loud failure rather than a silent wrong result.
- MC recompression remains absent despite the file's own TODO, so downstream MC products are the input file's, not recompressed against the regrown tracks.

### Author follow-ups (none blocking)
1. Narrow both `keep mu2e::*_RegrowKK*_*_*` to `keep mu2e::KalSeed*_RegrowKK*_*_*`, or say in the PR body that the `KKTrack` collection is intentionally persisted (finding 1).
2. Still open from the first pass, all non-gating: drop the two dead `RefCompressRecoMCs*` blocks and the two `nil` output lines; `RegrowSH : @local::Reconstruction.producers.RegrowSH`; hoist the shared `outputCommands`; one comment on `eidsel`.
3. Best practice reminder: please add validation evidence to the PR description — the command, the input dataset, and how many events survive each path.

Thanks for turning both blockers around quickly. The `BackToTracker` commit in particular is the piece that makes the three-stage chain make sense, and reading this file is what corrected my own misreading of #1917.

PR #567, opened 2026-08-14
Updates for ensembling

sophiemiddleton:
This pull request adds support for Radiative Muon Capture (RMC) 0-nucleon (0N) and 1-nucleon (1N) knockout processes to the normalization and event calculation framework. It introduces new normalization functions, updates the event calculation and template generation scripts to handle these processes, and adds corresponding tests. There are also minor updates to configuration and test files to support these changes.

**Normalization and Physics Calculation Enhancements:**

* Added new physics constants and functions for RMC 0N and 1N knockout processes, including spectrum integration and normalization functions (`rmc_0n_normalization`, `rmc_1n_normalization`, and supporting Plestid spectrum utilities) in `normalizations.py` [[1]](diffhunk://#diff-b67289f2b009431f1835dcaa17bf62a7c2c3e28e095119123833efa5483c3a39R61-R71) [[2]](diffhunk://#diff-b67289f2b009431f1835dcaa17bf62a7c2c3e28e095119123833efa5483c3a39L503-R696).
* Extended argument parsing and event calculation in `calculateEvents.py` to support new RMC 0N and 1N processes, including both internal and external conversion options [[1]](diffhunk://#diff-e2d5412ee13abe339502adb0ecae6bde3dc101cb0658a90f5fe026eda6397e78R2) [[2]](diffhunk://#diff-e2d5412ee13abe339502adb0ecae6bde3dc101cb0658a90f5fe026eda6397e78R47-R70) [[3]](diffhunk://#diff-e2d5412ee13abe339502adb0ecae6bde3dc101cb0658a90f5fe026eda6397e78R86-R87).

**Template Generation and Configuration:**

* Updated `make_template_fcl.py` to allow selective calculation of normalization for new RMC 0N/1N processes, with argument parsing and logic for these processes [[1]](diffhunk://#diff-3ee5f34e877d2d87639a7625db46afa9bd406561245c520ea43b092224821682L35-R76) [[2]](diffhunk://#diff-3ee5f34e877d2d87639a7625db46afa9bd406561245c520ea43b092224821682R244-R245).
* Modified FCL configuration to drop the new `mu2e::SpectrumConfig` product, keeping output files clean.

**Testing Improvements:**

* Added comprehensive unit tests for RMC 0N and 1N normalization functions, including both internal and external yields, and updated summary printouts in `test_normalizations.py` [[1]](diffhunk://#diff-ccfc545134c006a92ee27adf9536faeeb071adb2cf4387f88a9eb03b7b21c0c9R204-R275) [[2]](diffhunk://#diff-ccfc545134c006a92ee27adf9536faeeb071adb2cf4387f88a9eb03b7b21c0c9R311-R314) [[3]](diffhunk://#diff-ccfc545134c006a92ee27adf9536faeeb071adb2cf4387f88a9eb03b7b21c0c9R324-R327) [[4]](diffhunk://#diff-ccfc545134c006a92ee27adf9536faeeb071adb2cf4387f88a9eb03b7b21c0c9R341-R344) [[5]](diffhunk://#diff-ccfc545134c006a92ee27adf9536faeeb071adb2cf4387f88a9eb03b7b21c0c9R354-R357).
* Adjusted test setup to use more realistic spill times.

These changes collectively enable precise modeling and normalization of RMC 0N and 1N backgrounds, improving the physics reach and flexibility of the event calculation framework.

### DIFF
--- a/JobConfig/ensemble/python/calculateEvents.py
+++ b/JobConfig/ensemble/python/calculateEvents.py
@@ -1,4 +1,5 @@
 #! /usr/bin/env python
+import argparse
 from normalizations import *
 
 def main(args):
@@ -43,6 +44,30 @@ def main(args):
     if(args.prc == "RMC" and int(args.internal) == 0):
       Yield = rmc_normalization(float(args.livetime),  str(args.internal), float(args.rmcemin))
       print("ExternalRMC_yield=",Yield)
+    if(args.prc == "RMCN0External"):
+      Yield = rmc_0n_normalization(float(args.livetime), float(args.rmcn0emin), internal=0, run_mode=str(args.BB))
+      print("ExternalRMCN0_yield=",Yield)
+    if(args.prc == "RMCN0Internal"):
+      Yield = rmc_0n_normalization(float(args.livetime), float(args.rmcn0emin), internal=1, run_mode=str(args.BB))
+      print("InternalRMCN0_yield=",Yield)
+    if(args.prc == "RMCPhaseSpace0NExternal"):
+      Yield = rmc_0n_normalization(float(args.livetime), float(args.rmcn0emin), internal=0, run_mode=str(args.BB))
+      print("ExternalRMCPhaseSpace0N_yield=",Yield)
+    if(args.prc == "RMCPhaseSpace0NInternal"):
+      Yield = rmc_0n_normalization(float(args.livetime), float(args.rmcn0emin), internal=1, run_mode=str(args.BB))
+      print("InternalRMCPhaseSpace0N_yield=",Yield)
+    if(args.prc == "RMCN1External"):
+      Yield = rmc_1n_normalization(float(args.livetime), float(args.rmcn1emin), internal=0, run_mode=str(args.BB))
+      print("ExternalRMCN1_yield=",Yield)
+    if(args.prc == "RMCN1Internal"):
+      Yield = rmc_1n_normalization(float(args.livetime), float(args.rmcn1emin), internal=1, run_mode=str(args.BB))
+      print("InternalRMCN1_yield=",Yield)
+    if(args.prc == "RMCPhaseSpace1NExternal"):
+      Yield = rmc_1n_normalization(float(args.livetime), float(args.rmcn1emin), internal=0, run_mode=str(args.BB))
+      print("ExternalRMCPhaseSpace1N_yield=",Yield)
+    if(args.prc == "RMCPhaseSpace1NInternal"):
+      Yield = rmc_1n_normalization(float(args.livetime), float(args.rmcn1emin), internal=1, run_mode=str(args.BB))
+      print("InternalRMCPhaseSpace1N_yield=",Yield)
     if(args.prc == "IPAMichel"):
       Yield = ipaMichel_normalization(float(args.livetime), float(args.ipaemin), str(args.BB))
       print("IPAMichel_yield=",Yield)
@@ -58,6 +83,8 @@ if __name__ == '__main__':
     parser.add_argument("--ipaemin", help="min energy cut dio ipa")
     parser.add_argument("--rpcemin", help="rpcemin", default=0)
     parser.add_argument("--rmcemin", help="min energy cut rmc")
+    parser.add_argument("--rmcn0emin", help="min energy cut rmc 0N")
+    parser.add_argument("--rmcn1emin", help="min energy cut rmc 1N")
     parser.add_argument("--prc", help="process")
     parser.add_argument("--printpot", help="print pot", default="no")
     parser.add_argument("--tmin", help="tmin", default=0)
--- a/JobConfig/ensemble/python/normalizations.py
+++ b/JobConfig/ensemble/python/normalizations.py
@@ -58,6 +58,17 @@ rate = 1.0
 dutyfactor = 1.0
 total_pot = 0.
 
+# RMC 0N and 1N Physics Constants
+RMC_BR_MUON_CAPTURE = 0.609
+RMC_RATE_GT_57 = 1.41e-5  # RMC rate above 57 MeV, relative to OMC
+RMC_BR_0N_FRAC_GT_57 = 0.099  # BR(0 knockout | E > 57) / BR(RMC | E > 57)
+RMC_BR_1N_FRAC_GT_57 = 0.901  # BR(1 knockout | E > 57) / BR(RMC | E > 57)
+RMC_SPECTRUM_FRAC_0N_57 = 0.22887  # R(0 knockout | E > 57) / R(0 knockout)
+RMC_SPECTRUM_FRAC_1N_57 = 0.061620  # R(1 knockout | E > 57) / R(1 knockout)
+RMC_SPECTRUM_FRAC_0N_80 = 0.03319  # R(0 knockout | E > 80) / R(0 knockout)
+RMC_SPECTRUM_FRAC_1N_80 = 0.0013175  # R(1 knockout | E > 80) / R(1 knockout)
+RMC_INTERNAL_EXTERNAL_RATIO = 0.0069  # rho = BR(internal) / BR(external)
+
 #-------------------------------------------------------------------------------------#
 
 # --- Database Interaction ---
@@ -500,7 +511,189 @@ def rmc_normalization(on_spill_time, internal, e_min, k_max=90.1, run_mode='1BB'
         
     return base_physics_events
 
-# get IPA Michel normalization:
+#-------------------------------------------------------------------------------------#
+# Plestid spectrum functions for RMC 0N and 1N
+
+def plestid_integral(K_1, K_2, KMax, knockout):
+    """
+    Calculates the integral of the Plestid phase-space approximation spectrum
+    between two energy points K_1 and K_2.
+    
+    This implements the Plestid spectrum shape for RMC with different knockout modes.
+    
+    Args:
+        K_1 (float): Lower energy bound for integration (MeV).
+        K_2 (float): Upper energy bound for integration (MeV).
+        KMax (float): Maximum possible RMC energy (MeV).
+        knockout (int): Knockout mode (0 for 0N knockout, 1 for 1N knockout).
+    
+    Returns:
+        float: The integral of the spectrum between K_1 and K_2.
+    """
+    if KMax <= 0.0:
+        return 0.0
+    if knockout < 0:
+        return 0.0
+    
+    K_1 = max(0.0, min(KMax, K_1))
+    K_2 = max(0.0, min(KMax, K_2))
+    
+    if K_1 >= K_2:
+        return 0.0
+    
+    power = 2.0 + 1.5 * knockout
+    x_1 = K_1 / KMax
+    x_2 = K_2 / KMax
+    
+    val_1 = (x_1 - 1.0) * pow(1.0 - x_1, power) * (power * x_1 + x_1 + 1.0)
+    val_2 = (x_2 - 1.0) * pow(1.0 - x_2, power) * (power * x_2 + x_2 + 1.0)
+    
+    integral = val_2 - val_1
+    return integral
+
+
+def plestid_spectrum(energy, kmax, knockout):
+    """
+    Calculates the Plestid phase-space approximation spectrum value at a given energy.
+    
+    Args:
+        energy (float): Energy point to evaluate the spectrum (MeV).
+        kmax (float): Maximum possible RMC energy (MeV).
+        knockout (int): Knockout mode (0 for 0N knockout, 1 for 1N knockout).
+    
+    Returns:
+        float: The spectrum value at the given energy.
+    """
+    if energy <= 0.0 or energy >= kmax:
+        return 0.0
+    
+    power = 2.0 + 1.5 * knockout
+    norm = (power + 1.0) * (power + 2.0) / kmax
+    x = energy / kmax
+    p = norm * x * pow(1.0 - x, power)
+    
+    return p
+
+
+def rmc_0n_normalization(on_spill_time, e_min, internal=1, run_mode='1BB'):
+    """
+    Calculates the expected number of RMC 0-nucleon knockout (0N) events
+    above a given energy threshold.
+    
+    Uses the Plestid phase-space approximation spectrum shape.
+    
+    Args:
+        on_spill_time (float): Time the beam was on spill (seconds).
+        e_min (float): Minimum energy threshold for the spectrum cut (MeV).
+        internal (int/bool): Flag (1 or 0) to include internal conversion scaling.
+                            Defaults to 1.
+        run_mode (str): The operational mode ('1BB' or '2BB'). Defaults to '1BB'.
+    
+    Returns:
+        float: The expected number of RMC 0N physics events passing the cuts.
+    """
+    # 1. Calculate total Protons on Target (POT)
+    total_pot = get_pot(on_spill_time, run_mode)
+    
+    # 2. Determine the spectrum fraction to use based on energy threshold
+    # Default: fraction from E > 57 MeV
+    e_threshold = 57.0
+    R_spectrum = RMC_SPECTRUM_FRAC_0N_57
+    
+    # If threshold is higher, use interpolated fraction
+    # For E > 80 MeV: use the E > 80 spectrum fraction
+    if float(e_min) > 75.0:
+        R_spectrum = RMC_SPECTRUM_FRAC_0N_80 / RMC_SPECTRUM_FRAC_0N_57
+    
+    # 3. Calculate the branching ratio for 0N events above the energy threshold
+    br_0n_above_emin = (
+        RMC_BR_MUON_CAPTURE *
+        RMC_RATE_GT_57 *
+        RMC_BR_0N_FRAC_GT_57 *
+        R_spectrum
+    )
+    
+    # 4. Calculate base physics events
+    # Note: br_0n_above_emin already includes RMC_BR_MUON_CAPTURE, so we do NOT multiply by CAPTURES_PER_STOPPED_MUON
+    base_physics_events = (
+        total_pot *
+        target_stopped_muons_per_pot *
+        br_0n_above_emin
+    )
+    
+    # 5. Apply internal conversion scaling if requested
+    is_internal_conversion = bool(int(internal))
+    
+    if is_internal_conversion:
+        if VERBOSE:
+            print("RMC_0N_emin=", e_min)
+            print("RMC_0N_spectrum_frac=", R_spectrum)
+            print("RMC_0N_BR=", br_0n_above_emin)
+        
+        base_physics_events *= RMC_INTERNAL_EXTERNAL_RATIO
+    
+    return base_physics_events
+
+
+def rmc_1n_normalization(on_spill_time, e_min, internal=1, run_mode='1BB'):
+    """
+    Calculates the expected number of RMC 1-nucleon knockout (1N) events
+    above a given energy threshold.
+    
+    Uses the Plestid phase-space approximation spectrum shape.
+    
+    Args:
+        on_spill_time (float): Time the beam was on spill (seconds).
+        e_min (float): Minimum energy threshold for the spectrum cut (MeV).
+        internal (int/bool): Flag (1 or 0) to include internal conversion scaling.
+                            Defaults to 1.
+        run_mode (str): The operational mode ('1BB' or '2BB'). Defaults to '1BB'.
+    
+    Returns:
+        float: The expected number of RMC 1N physics events passing the cuts.
+    """
+    # 1. Calculate total Protons on Target (POT)
+    total_pot = get_pot(on_spill_time, run_mode)
+    
+    # 2. Determine the spectrum fraction to use based on energy threshold
+    # Default: fraction from E > 57 MeV
+    e_threshold = 57.0
+    R_spectrum = RMC_SPECTRUM_FRAC_1N_57
+    
+    # If threshold is higher, use interpolated fraction
+    # For E > 80 MeV: use the E > 80 spectrum fraction
+    if float(e_min) > 75.0:
+        R_spectrum = RMC_SPECTRUM_FRAC_1N_80 / RMC_SPECTRUM_FRAC_1N_57
+    
+    # 3. Calculate the branching ratio for 1N events above the energy threshold
+    br_1n_above_emin = (
+        RMC_BR_MUON_CAPTURE *
+        RMC_RATE_GT_57 *
+        RMC_BR_1N_FRAC_GT_57 *
+        R_spectrum
+    )
+    
+    # 4. Calculate base physics events
+    # Note: br_1n_above_emin already includes RMC_BR_MUON_CAPTURE, so we do NOT multiply by CAPTURES_PER_STOPPED_MUON
+    base_physics_events = (
+        total_pot *
+        target_stopped_muons_per_pot *
+        br_1n_above_emin
+    )
+    
+    # 5. Apply internal conversion scaling if requested
+    is_internal_conversion = bool(int(internal))
+    
+    if is_internal_conversion:
+        if VERBOSE:
+            print("RMC_1N_emin=", e_min)
+            print("RMC_1N_spectrum_frac=", R_spectrum)
+            print("RMC_1N_BR=", br_1n_above_emin)
+        
+        base_physics_events *= RMC_INTERNAL_EXTERNAL_RATIO
+    
+    return base_physics_events
+
 def ipaMichel_normalization(on_spill_time, ipa_de_min, run_mode='1BB'):
     """
     Calculates the expected number of IPA (Incoming Particle Decay After Stopping)
--- a/JobConfig/ensemble/fcl/prolog.fcl
+++ b/JobConfig/ensemble/fcl/prolog.fcl
@@ -6,7 +6,7 @@ Ensemble : {
           "keep mu2e::CosmicLivetime_*_*_*",
           "drop *_genCounter_*_*",
           "drop *_protonBunchIntensity_*_*",
-          "drop *_mu2e::SpectrumConfig_*_*"]
+          "drop mu2e::SpectrumConfig_*_*_*"]
     OutputCommandsMC : [   "keep *_*_*_*"]
     OutputCommandsData : [   "drop *_*_*_*",
           "keep mu2e::KalSeeds_*_*_*",
--- a/JobConfig/ensemble/fcl/prolog.fcl
+++ b/JobConfig/ensemble/fcl/prolog.fcl
@@ -5,7 +5,8 @@ Ensemble : {
     InputCommands : [ "keep *", 
           "keep mu2e::CosmicLivetime_*_*_*",
           "drop *_genCounter_*_*",
-          "drop *_protonBunchIntensity_*_*"]
+          "drop *_protonBunchIntensity_*_*",
+          "drop *_mu2e::SpectrumConfig_*_*"]
     OutputCommandsMC : [   "keep *_*_*_*"]
     OutputCommandsData : [   "drop *_*_*_*",
           "keep mu2e::KalSeeds_*_*_*",
--- a/JobConfig/ensemble/python/make_template_fcl.py
+++ b/JobConfig/ensemble/python/make_template_fcl.py
@@ -32,17 +32,48 @@ def main(args):
 
   ROOT.gRandom.SetSeed(0)
 
-  # extract normalization of each background/signal process:
-  norms = {
-          "CRYCosmic": cry_onspill_normalization(livetime, args.BB),
-          "CORSIKACosmic": corsika_onspill_normalization(livetime, args.BB),
-          "DIO": dio_normalization(livetime, dioemin, args.BB),
-          "RPCInternal": rpc_normalization(livetime, args.tmin, 1, args.rpcemin, args.BB),
-          "RPCExternal": rpc_normalization(livetime, args.tmin, 0, args.rpcemin, args.BB),
-          "RMCInternal": rmc_normalization(livetime, 1, args.rmcemin, args.rmckmax, args.BB),
-          "RMCExternal": rmc_normalization(livetime, 0, args.rmcemin, args.rmckmax, args.BB),
-          "IPAMichel": ipaMichel_normalization(livetime, args.ipaemin, args.BB)
-          }
+  # Convert args.prc into a set for fast lookup
+  requested_processes = set(args.prc)
+  
+  # Initialize an empty dictionary
+  norms = {}
+
+  # Only call functions and define keys if they are in the requested processes
+  if "CRYCosmic" in requested_processes:
+      norms["CRYCosmic"] = cry_onspill_normalization(livetime, args.BB)
+      
+  if "CORSIKACosmic" in requested_processes:
+      norms["CORSIKACosmic"] = corsika_onspill_normalization(livetime, args.BB)
+      
+  if "DIO" in requested_processes:
+      norms["DIO"] = dio_normalization(livetime, dioemin, args.BB)
+      
+  if "RPCInternal" in requested_processes:
+      norms["RPCInternal"] = rpc_normalization(livetime, tmin, 1, args.rpcemin, args.BB)
+      
+  if "RPCExternal" in requested_processes:
+      norms["RPCExternal"] = rpc_normalization(livetime, tmin, 0, args.rpcemin, args.BB)
+      
+  if "RMCInternal" in requested_processes:
+      norms["RMCInternal"] = rmc_normalization(livetime, 1, args.rmcemin, args.rmckmax, args.BB)
+      
+  if "RMCExternal" in requested_processes:
+      norms["RMCExternal"] = rmc_normalization(livetime, 0, args.rmcemin, args.rmckmax, args.BB)
+      
+  if "RMCN0External" in requested_processes or "RMCPhaseSpace0NExternal" in requested_processes:
+      norms["RMCN0External"] = norms["RMCPhaseSpace0NExternal"] = rmc_0n_normalization(livetime, args.rmcn0emin, internal=0, run_mode=args.BB)
+      
+  if "RMCN0Internal" in requested_processes or "RMCPhaseSpace0NInternal" in requested_processes:
+      norms["RMCN0Internal"] = norms["RMCPhaseSpace0NInternal"] = rmc_0n_normalization(livetime, args.rmcn0emin, internal=1, run_mode=args.BB)
+      
+  if "RMCN1External" in requested_processes or "RMCPhaseSpace1NExternal" in requested_processes:
+      norms["RMCN1External"] = norms["RMCPhaseSpace1NExternal"] = rmc_1n_normalization(livetime, args.rmcn1emin, internal=0, run_mode=args.BB)
+      
+  if "RMCN1Internal" in requested_processes or "RMCPhaseSpace1NInternal" in requested_processes:
+      norms["RMCN1Internal"] = norms["RMCPhaseSpace1NInternal"] = rmc_1n_normalization(livetime, args.rmcn1emin, internal=1, run_mode=args.BB)
+      
+  if "IPAMichel" in requested_processes:
+      norms["IPAMichel"] = ipaMichel_normalization(livetime, args.ipaemin, args.BB)
 
   starting_event_num = {}
   max_possible_events = {}
@@ -210,6 +241,8 @@ if __name__ == "__main__":
     parser.add_argument("--rpcemin", help="min energy cut rpc")
     parser.add_argument("--ipaemin", help="min energy cut ipa")
     parser.add_argument("--rmcemin", help="min energy cut rmc")
+    parser.add_argument("--rmcn0emin", help="min energy cut rmc 0N")
+    parser.add_argument("--rmcn1emin", help="min energy cut rmc 1N")
     parser.add_argument("--rmckmax", help="kmax theory value")
     parser.add_argument("--run", help="run number")
     parser.add_argument("--samplingseed", help="samplingseed")
--- a/JobConfig/ensemble/python/test_normalizations.py
+++ b/JobConfig/ensemble/python/test_normalizations.py
@@ -25,8 +25,8 @@ class TestNormalizations(unittest.TestCase):
             os.environ['MUSE_WORK_DIR'] = '/exp/mu2e/app/users/sophie/newOffline'
 
         # Test livetimes from MDS3 (cosmic livetime)
-        cls.on_spill_time_1BB = 4.4e6  # seconds
-        cls.on_spill_time_2BB = 4.4e6  # seconds
+        cls.on_spill_time_1BB = 4.4e6 * (88/496) # seconds
+        cls.on_spill_time_2BB = 4.4e6 * (88/496)  # seconds
         
         cls.tst_1BB_cycle = normalizations.get_pot(cls.on_spill_time_1BB, run_mode='1BB',printout=False,method='cycle')
         cls.tst_2BB_cycle = normalizations.get_pot(cls.on_spill_time_2BB, run_mode='2BB',printout=False,method='cycle')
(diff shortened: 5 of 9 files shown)


JobConfig/ensemble/python/calculateEvents.py
```diff
@@ -43,6 +44,30 @@ def main(args):
     if(args.prc == "RMC" and int(args.internal) == 0):
       Yield = rmc_normalization(float(args.livetime),  str(args.internal), float(args.rmcemin))
       print("ExternalRMC_yield=",Yield)
+    if(args.prc == "RMCN0External"):
+      Yield = rmc_0n_normalization(float(args.livetime), float(args.rmcn0emin), internal=0, run_mode=str(args.BB))
+      print("ExternalRMCN0_yield=",Yield)
+    if(args.prc == "RMCN0Internal"):
+      Yield = rmc_0n_normalization(float(args.livetime), float(args.rmcn0emin), internal=1, run_mode=str(args.BB))
+      print("InternalRMCN0_yield=",Yield)
+    if(args.prc == "RMCPhaseSpace0NExternal"):
```
michaelmackenzie, line 43: Why both RMCN0External and RMCPhaseSpace0NExternal (for example)?

JobConfig/ensemble/python/calculateEvents.py
```diff
@@ -58,6 +83,8 @@ def main(args):
     parser.add_argument("--ipaemin", help="min energy cut dio ipa")
     parser.add_argument("--rpcemin", help="rpcemin", default=0)
     parser.add_argument("--rmcemin", help="min energy cut rmc")
+    parser.add_argument("--rmcn0emin", help="min energy cut rmc 0N")
```
michaelmackenzie, line 70: This assumes the internal and external photon energies have the same cuts, right? I think this is fine, just checking it's intentional.

JobConfig/ensemble/python/normalizations.py
```diff
@@ -500,7 +511,189 @@ def rmc_normalization(on_spill_time, internal, e_min, k_max=90.1, run_mode='1BB'
...
+    total_pot = get_pot(on_spill_time, run_mode)
+    
+    # 2. Determine the spectrum fraction to use based on energy threshold
+    # Default: fraction from E > 57 MeV
+    e_threshold = 57.0
+    R_spectrum = RMC_SPECTRUM_FRAC_0N_57
+    
+    # If threshold is higher, use interpolated fraction
+    # For E > 80 MeV: use the E > 80 spectrum fraction
+    if float(e_min) > 75.0:
```
michaelmackenzie: This is set at 75 instead of 80 MeV. And why use hard-coded R_spectrum value instead of doing the integral of the plestid spectrum from emin to kmax? Will this hold for arbitrary emin values?

JobConfig/ensemble/python/normalizations.py
```diff
@@ -500,7 +511,189 @@ def rmc_normalization(on_spill_time, internal, e_min, k_max=90.1, run_mode='1BB'
...
+    total_pot = get_pot(on_spill_time, run_mode)
+    
+    # 2. Determine the spectrum fraction to use based on energy threshold
+    # Default: fraction from E > 57 MeV
+    e_threshold = 57.0
+    R_spectrum = RMC_SPECTRUM_FRAC_1N_57
+    
+    # If threshold is higher, use interpolated fraction
+    # For E > 80 MeV: use the E > 80 spectrum fraction
+    if float(e_min) > 75.0:
```
michaelmackenzie: Same comment here as above

michaelmackenzie (changes requested):
Overall I think this looks great, thank you for making these updates! My only request/comment is to ensure the RMC normalization is valid for arbitrary values of emin (which may already be true).

JobConfig/ensemble/python/calculateEvents.py
sophiemiddleton: I ran my test_normalization with Run-1A and the yield was the same as what you sent me from your calculations. These are just nicknames really, but yes they should match the full name I guess

JobConfig/ensemble/python/calculateEvents.py
sophiemiddleton: yes it assumes, I dont want too many parameters. If in future they differ I can add more

JobConfig/ensemble/python/normalizations.py
sophiemiddleton: good spot, I will check this

oksuzian (commented):
### PR Review Summary — Mu2e/Production#567

Reviewed at head `274b68d8c205c90201c45c9381bab6b67eb0226b`. First pass; @michaelmackenzie's CHANGES_REQUESTED at this same head is carried forward below.

**Decision**
- 🟡 comment only

Both S1s are in the normalization arithmetic rather than the plumbing, and
neither fires at the committed defaults (Stage1 sets both thresholds to 80 MeV) —
hence comment rather than block. @michaelmackenzie's request is still open, and
his instinct was right; the numbers are below.

**Scope understood**
- Adds RMC 0-nucleon and 1-nucleon knockout normalizations (`rmc_0n_normalization`,
  `rmc_1n_normalization`) plus Plestid spectrum helpers, and wires them through
  `calculateEvents.py`, `make_template_fcl.py` and Stages 1/2/3.
- The ensemble's plain `RMCInternal`/`RMCExternal` entries are replaced by the
  0N/1N pair, not supplemented — `BR_0N_FRAC + BR_1N_FRAC = 0.099 + 0.901 = 1.0`,
  so this is the same RMC rate decomposed by knockout mode, and no background is
  dropped. Worth one line in the description, since "adds support" reads as
  additive.
- Also carries an unrelated `DBVERSION` parsing fix in `Stage5_signal.sh` and a
  test-livetime change.

**Carried forward from @michaelmackenzie's review** (state: CHANGES_REQUESTED, no commits since)

- *"ensure the RMC normalization is valid for arbitrary values of emin"* and the
  two inline comments on `normalizations.py:605`/`:665` — **UNADDRESSED**, and
  finding 1 below gives the exact failure.
- *"Why both RMCN0External and RMCPhaseSpace0NExternal?"* — **UNADDRESSED** in
  code; the author has agreed to unify. See finding 5.
- *"This assumes the internal and external photon energies have the same cuts"* —
  **CLEARED** by the author's answer that it is intentional.

**Findings**

1. 🟠 [S1] For `e_min ≤ 75` the spectrum factor is an absolute fraction where a
   relative one is required, under-predicting by 4.4× (0N) and 16× (1N)
   - Evidence: `normalizations.py:601-606`. The prefactor
     `RMC_BR_MUON_CAPTURE * RMC_RATE_GT_57 * RMC_BR_0N_FRAC_GT_57` is already
     `BR(0N, E > 57)` per stopped muon, so `R_spectrum` has to be
     `R(>e_min)/R(>57)`. The `e_min > 75` branch computes exactly that —
     `FRAC_0N_80 / FRAC_0N_57 = 0.03319/0.22887 = 0.14501`. The default branch
     instead uses `FRAC_0N_57 = 0.22887`, whose denominator is the *whole*
     spectrum, applying a spectrum cut that the prefactor already applied. At
     `e_min = 57` the correct factor is `1.0`, so the result is low by
     `1/0.22887 = 4.37×`; for 1N, by `1/0.061620 = 16.2×`.
   - Impact: any `--rmcn0emin`/`--rmcn1emin` at or below 75 MeV silently yields a
     too-small RMC background. Stage1 hardcodes 80, so today's default path takes
     the correct branch — this is a latent wrong-physics path, not a live one.
     The 75.0 cut is itself discontinuous: 74 MeV gets the E>57 normalization and
     76 MeV gets the E>80 one, with nothing in between.
   - Suggested fix: the machinery to do this exactly is already in the PR, unused.
     `plestid_integral` is the correctly normalized CDF (I verified its
     antiderivative integrates the `plestid_spectrum` shape to 1 over `[0, kmax]`),
     so the whole branch collapses to

     ```python
     R_spectrum = (plestid_integral(e_min, kmax, kmax, knockout) /
                   plestid_integral(57.0,  kmax, kmax, knockout))
     ```

     and it then holds for arbitrary `e_min`, which is precisely what
     @michaelmackenzie asked for. The two `kmax` values are recoverable from your
     own constants — see finding 3.

2. 🟠 [S1] Three constants are re-declared locally, and one disagrees with the
   cited shared value
   - Evidence: `normalizations.py:62-70` against `constants.py`, which this same
     module already imports at line 14:
     - `RMC_RATE_GT_57 = 1.41e-5` vs `RMC_GT_57_PER_CAPTURE = 1.43e-5`
       — **different values for the same quantity**, and the shared one carries a
       source (`Phys. Rev. C 59, 2853 (1999)`) while the new one carries none.
     - `RMC_BR_MUON_CAPTURE = 0.609` duplicates `CAPTURES_PER_STOPPED_MUON = 0.609`.
     - `RMC_INTERNAL_EXTERNAL_RATIO = 0.0069` duplicates `INTERNAL_PER_RMC = 0.00690`.
   - Impact: `rmc_normalization` (line 490-495) uses the imported constants while
     `rmc_0n_normalization`/`rmc_1n_normalization` use the local copies, so the
     old and new RMC paths now disagree by 1.4% on the same rate. Whichever value
     is right, two homes for it means a future correction reaches only one.
   - Suggested fix: delete all three locals and use the imported names. If
     `1.41e-5` is a deliberate update, change it in `constants.py` with its source
     and let `rmc_normalization` move with it.

3. 🟡 [S2] `plestid_integral` and `plestid_spectrum` are added but never called
   - Evidence: `normalizations.py:517-575`, 59 lines; the only other occurrences
     of "Plestid" in the file are docstring prose in the two normalization
     functions, which use the hardcoded constants instead.
   - Worth keeping rather than deleting, because the four hardcoded fractions
     *are* this function: solving `plestid_integral(e, kmax, kmax, knockout)` for
     the `kmax` that reproduces each one gives **101.866 MeV** from `FRAC_0N_57`
     and **101.865 MeV** from `FRAC_0N_80`, and **95.449 MeV** from both 1N
     values — self-consistent to 0.001 MeV within each mode, and physically
     sensible (1N is lower by roughly the neutron separation energy). Those two
     numbers are the missing input; name them, and finding 1's fix is a two-line
     change. Note neither is the `RMC_kmax=90.1` that Stage1 still writes and
     Stage2 still passes as `--rmckmax`.

4. 🟡 [S2] The four new tests assert only `> 0`, and CI does not run them
   - Evidence: `test_normalizations.py:204-275` — each new test ends in
     `assertGreater(..., 0)`. A 4× or 16× normalization error passes. FNALbuild's
     table for build 658 lists 16 jobs (build, ceSimReco, ceMix, rootOverlaps,
     g4surfaceCheck, …) and no python test target, so nothing runs this file
     automatically either.
   - Suggested fix: pin one expected yield per mode to a few significant figures.
     Given finding 1, a second case at `e_min = 57` would have caught it.

5. 🟡 [S2] `on_spill_time` is scaled by an uncited `88/496`, identically for 1BB and 2BB
   - Evidence: `test_normalizations.py:28-29` — `4.4e6 * (88/496)` = 0.1774, applied
     to both modes. `constants.py` gives `ONEBB_DF = 0.323` and `TWOBB_DF = 0.246`;
     0.1774 is neither, and the two modes have different duty factors, so one
     shared literal cannot be right for both.
   - Suggested fix: use the named duty-factor constants, or comment where 88/496
     comes from.

6. 🟡 [S2] `Stage2_build_sampler.sh`: one dataset is validated on disk and then
   listed from tape, and the new file-list check only warns
   - Evidence: line 214 validates every dataset with
     `mu2eDatasetFileList "$dataset_name" --disk`, but the RPCInternal list at
     line 287 is built with `--tape` — alone among the six list-building calls,
     the other five pass no flag. Separately, `check_file_lists` (line 265, used
     at 301 and 363) prints
     `⚠️ WARNING: Some file lists may be empty or incomplete` and execution
     continues; an empty list surfaces much later as an `IndexError` at
     `make_template_fcl.py:201` or a `ZeroDivisionError` at line 164.
   - Suggested fix: drop the stray `--tape`, and make the file-list check `exit 1`.
     A check that warns and proceeds reads like a gate but is not one.

7. ⚪ [S3] Three small dead/contradictory items
   - `e_threshold = 57.0` is assigned and never read, in both new functions
     (`normalizations.py:600`, `:660`).
   - The alias assignments `norms["RMCN0External"] = norms["RMCPhaseSpace0NExternal"] = ...`
     (`make_template_fcl.py:63-73`) only ever populate keys nothing reads — the
     consumer loop is `for signal in args.prc` (line 85) and Stage2 passes only the
     `RMCPhaseSpace*` names. This is the code side of @michaelmackenzie's
     "why both names?"; dropping the short aliases resolves both.
   - `Stage1_initiate_ensemble.sh:24-25` — `INCLUDE_RMCN0=1 # ... (default: no)`.
     The comment contradicts the value; both modes are on by default, so
     `--rmcn0`/`--rmcn1` are opt-out, not opt-in.

- 🟢 The `Stage5_signal.sh` `DBVERSION` change is a real fix, not a rewrite. The
  old `$(IFS='_'; echo "${PARTS[@]:2}")` does not join on `IFS` — only `"${arr[*]}"`
  does, `"${arr[@]}"` expands to separate words that `echo` joins with spaces. Run
  on `MDC2025_best_v1_3` it returns `v1 3`; the new explicit loop returns `v1_3`.
- 🟢 `make_template_fcl.py:52,55` switching `args.tmin` → `tmin` is safe and is
  itself a fix: `tmin` is defined at line 26 as `float(args.tmin)`, so
  `rpc_normalization` now receives a float rather than the raw string.
- 🟢 CI is green at this head — FNALbuild build 658, all 16 rows ✅, including
  `rootOverlaps`, `g4surfaceCheck`, clang-tidy (0/0) and FIXME/TODO (0/0).

**Validation check**
- Build/tests run: CI green at `274b68d8` (build 658). The new python tests are
  not part of that suite; I did not execute them. My own checks were arithmetic
  on the constants and a bash reproduction of the `DBVERSION` parse.
- Config contract check: partial. The Stage1 → Stage2 variable handoff
  (`RMC_N0_emin`/`RMC_N1_emin` gating dataset lists, `--prc`, and sampling
  options) is internally consistent; I did not run the stages end to end.
- Cross-repo consistency: n/a — no Offline or trig-config interface is touched.

**Residual risk**
- I did not verify that the datasets the new Stage2 branches expect
  (`dts.mu2e.RMCPhaseSpace{0,1}N{Internal,External}.MDC2025{au,at}.art`) exist with
  enough files, nor that `RMCVERSIONINT=au`/`RMCVERSIONEXT=at` are the intended
  versions — that split is asymmetric and unexplained in the diff.
- The physics inputs themselves (0.099/0.901 knockout split, and whether the
  Plestid phase-space shape is the right model here) are outside what I can check;
  I verified only internal consistency and the arithmetic.

**Author follow-ups**
1. Fix the `e_min ≤ 75` branch (finding 1) — using `plestid_integral` answers
   @michaelmackenzie's arbitrary-`e_min` request in the same change. The two `kmax`
   values you need are 101.87 MeV (0N) and 95.45 MeV (1N), recovered from your own
   constants.
2. Reconcile `1.41e-5` against the cited `RMC_GT_57_PER_CAPTURE = 1.43e-5` and drop
   the three duplicate constants (finding 2).
3. Pin at least one expected yield in the new tests, and add an `e_min = 57` case
   (finding 4).
4. Drop the stray `--tape` and make the file-list check exit non-zero (finding 6).

sophiemiddleton:
I have now pushed updates and fixes to your responses

JobConfig/ensemble/python/normalizations.py
```diff
@@ -118,6 +144,110 @@ def set_verbose(verbose=True):
...
+    frac_0_1n  = plestid_integral(0.0, RMC_KMAX_1N, RMC_KMAX_1N, 1)
+    frac_57_0n = plestid_integral(57.0, RMC_KMAX_0N, RMC_KMAX_0N, 0)
+    frac_57_1n = plestid_integral(57.0, RMC_KMAX_1N, RMC_KMAX_1N, 1)
+    frac_80_0n = plestid_integral(80.0, RMC_KMAX_0N, RMC_KMAX_0N, 0)
+    frac_80_1n = plestid_integral(80.0, RMC_KMAX_1N, RMC_KMAX_1N, 1)
+    
+    # Compute ratios: integral above threshold / integral from 0 to K_max
+    RMC_SPECTRUM_FRAC_0N_57 = frac_57_0n / frac_0_0n if frac_0_0n != 0 else 0.0
+    RMC_SPECTRUM_FRAC_1N_57 = frac_57_1n / frac_0_1n if frac_0_1n != 0 else 0.0
+    RMC_SPECTRUM_FRAC_0N_80 = frac_80_0n / frac_0_0n if frac_0_0n != 0 else 0.0
```
michaelmackenzie, line 221: I don't think the 80 MeV values are needed here, if `emin` is an input then in the RMC normalization it should evaluate the fraction above the input `emin` using the integral function directly

JobConfig/ensemble/python/normalizations.py
```diff
@@ -500,7 +630,126 @@ def rmc_normalization(on_spill_time, internal, e_min, k_max=90.1, run_mode='1BB'
...
+    total_pot = get_pot(on_spill_time, run_mode)
+    
+    # 2. Determine the spectrum fraction to use based on energy threshold
+    # Default: fraction from E > 57 MeV
+    e_threshold = 57.0
+    R_spectrum = RMC_SPECTRUM_FRAC_0N_57
+    
+    # If threshold is higher, use interpolated fraction
+    # For E > 80 MeV: use the E > 80 spectrum fraction
+    if float(e_min) >80.0:
```
michaelmackenzie: For all values of `e_min` it should evaluate the fraction above `e_min` and use `R_spectrum = frac_above_e_min / frac_above_57`

JobConfig/ensemble/python/normalizations.py
```diff
@@ -500,7 +630,126 @@ def rmc_normalization(on_spill_time, internal, e_min, k_max=90.1, run_mode='1BB'
...
+    total_pot = get_pot(on_spill_time, run_mode)
+    
+    # 2. Determine the spectrum fraction to use based on energy threshold
+    # Default: fraction from E > 57 MeV
+    e_threshold = 57.0
+    R_spectrum = RMC_SPECTRUM_FRAC_1N_57
+    
+    # If threshold is higher, use interpolated fraction
+    # For E > 80 MeV: use the E > 80 spectrum fraction
+    if float(e_min) >80.0:
```
michaelmackenzie: Same comment as above

michaelmackenzie (changes requested):

sophiemiddleton:
I think this is now ready for a merge.

JobConfig/ensemble/python/normalizations.py
```diff
@@ -58,6 +58,32 @@ def set_verbose(verbose=True):
...
+
+# RMC Spectrum Fractions - from experimental/theoretical physics literature
+# These represent: R(* knockout | E > threshold) / R(* knockout) 
+# i.e., the fraction of the full spectrum above the given energy threshold
+# Used to scale branching ratios as a function of energy cut:
+#   BR(* knockout | E > E_min) = BR(muon capture) * RMC_RATE_GT_57 * BR(* frac | E > 57) * (R_*_E_min / R_*_57)
+# Note: These are the RAW spectrum fractions (not pre-multiplied by RMC_BR_MUON_CAPTURE)
+RMC_SPECTRUM_FRAC_0N_57 = 0.22887  # Fraction of 0-knockout spectrum above 57 MeV
+RMC_SPECTRUM_FRAC_1N_57 = 0.061620  # Fraction of 1-knockout spectrum above 57 MeV
+RMC_SPECTRUM_FRAC_0N_80 = 0.03319  # Fraction of 0-knockout spectrum above 80 MeV
```
michaelmackenzie: I don't think these 80 MeV versions are used anymore

JobConfig/ensemble/python/normalizations.py
```diff
@@ -118,6 +144,120 @@ def set_verbose(verbose=True):
...
+    """
+    Compute RMC spectrum fractions dynamically using plestid_integral.
+    
+    Uses the correct K_max values for each knockout mode:
+    - kmax_0n = 101.8667 MeV (0-nucleon knockout endpoint on Al-27)
+    - kmax_1n = 95.4489 MeV (1-nucleon knockout endpoint on Al-27)
+    
+    These fractions represent: R(* knockout | E > threshold) / R(* knockout)
+    """
+    global RMC_SPECTRUM_FRAC_0N_57, RMC_SPECTRUM_FRAC_1N_57
```
michaelmackenzie, line 207: Are these values used anywhere? Is this function still needed?

michaelmackenzie (approved):
This looks good to me. I just had a couple of comments about no longer used code, but these are just comments and not required

JobConfig/ensemble/python/normalizations.py
sophiemiddleton: thanks, I thnk I was using these for validation

oksuzian (changes requested):
### PR Review Summary — Mu2e/Production#567

Reviewed at head `3c13c2828abb36da5f001b38079dc3568b55b305`. Re-review of `274b68d8`.

**Decision**
- 🔴 request changes

The RMC work is now right, and I verified it numerically rather than taking the
commit message for it — the arbitrary-`e_min` fix @michaelmackenzie asked for is
in, and it reproduces the production default to better than 0.02%. What blocks is
something else the delta carries: one line was deleted from `rpc_normalization`,
which multiplies the RPC background by 6.2 and is mentioned nowhere in the PR
description, the commit messages, or any review thread.

I note @michaelmackenzie approved at this head on 2026-08-20 with comments he
explicitly called non-blocking. His review was of the RMC normalization, and I
agree with it; finding 1 below is in a different function that the RMC discussion
never touched.

**Scope of this pass**
- `274b68d8..3c13c282`, four commits ("fixed norms" ×2, "made emin an input arf"
  ×2), across `normalizations.py` (+140/−89), `calculateEvents.py` (−16) and a new
  `test_rmc_fractions.py` (+76).

---

**Findings**

1. 🔴 [S0] **`filter_efficiency` was deleted from the RPC normalization, raising the RPC yield by a factor of 6.2, in a PR about RMC.**
   - Evidence: `normalizations.py:535` still computes
     `filter_efficiency = float(num_pion_filters) / float(num_pion_stops)`, but the
     product at `:538-544` no longer contains it. On `main` the same product does
     — `normalizations.py:401` there is the `filter_efficiency *` line this delta
     removes. The variable is now assigned and never read.
   - Impact: with the values the file's own comments record for the table it reads
     (`num_pion_stops = 41324703` at `:131`, `num_pion_filters = 6634478` at
     `:133`), `filter_efficiency = 0.1605`, so every RPC yield this module produces
     — internal and external, both BB modes — comes out **6.23× larger** than
     before. `rpc_normalization` is live: `make_template_fcl.py` calls it for the
     ensemble. Nothing in the PR description, the four commit messages, or any
     review thread mentions RPC.
   - Suggested fix: if this is deliberate — say, because the filter efficiency is
     already folded into `target_stopped_pions_per_pot` and was double-counted —
     then state it in the description, delete the now-dead `filter_efficiency`
     line, and split it into its own PR so the RPC change is reviewed as an RPC
     change. If it is not deliberate, restore the line. The dead variable left
     behind is what makes me think it is the second.

2. 🟠 [S1] **Carried over: three constants are still re-declared locally, and one still disagrees with the cited shared value.**
   `normalizations.py:62-63` still has `RMC_BR_MUON_CAPTURE = 0.609` and
   `RMC_RATE_GT_57 = 1.41e-5`, against `constants.py:9`
   `CAPTURES_PER_STOPPED_MUON = 0.609` and `constants.py:11`
   `RMC_GT_57_PER_CAPTURE = 1.43e-5 # Source: Phys. Rev. C 59, 2853 (1999)`;
   `RMC_INTERNAL_EXTERNAL_RATIO = 0.0069` still duplicates
   `constants.py:22 INTERNAL_PER_RMC = 0.00690`. The old `rmc_normalization` uses
   the imported names and the new 0N/1N functions use the locals, so the two RMC
   paths still disagree by 1.4% on the same rate, and only one of the two values
   carries a source. Delete the locals, or move `1.41e-5` into `constants.py` with
   its citation.

3. 🟡 [S2] **The plain-RMC branch in `calculateEvents.py` now computes a yield and prints nothing, and its external twin was deleted.**
   `calculateEvents.py:41-42` is `if(args.prc == "RMC" and int(args.internal) == 1):
   Yield = rmc_normalization(...)` with no `print`, where every other branch in the
   function prints a `<Name>_yield=` line for the shell to parse; the
   `int(args.internal) == 0` branch is gone entirely. The only remaining caller is
   `scripts/old/Stage1_makeinputs.sh:110,112`, which would now append nothing to
   `${TAG}.txt` instead of failing. Either finish the removal — drop both branches,
   the way the `RMCN0*`/`RMCN1*` aliases were dropped — or restore the print.

4. 🟡 [S2] **`test_rmc_fractions.py` is not a test and duplicates the function it checks.**
   The new 76-line file re-declares `plestid_integral` verbatim at `:11-33` rather
   than importing it from `normalizations.py`, so it validates its own copy and
   will silently stop tracking the real one. It contains no assertion — it computes
   four numbers, prints a comparison table and exits 0 regardless — while the
   `test_` prefix means unittest and pytest discovery will collect it and record a
   pass. Import `plestid_integral` and assert the four fractions against the
   literature values, or move it out of `test_*` and call it a script.

5. 🟡 [S2] **The four `RMC_SPECTRUM_FRAC_*` constants and the function that computes them are now entirely dead** — this is @michaelmackenzie's `:79` and `:222` comments, which he marked non-blocking; I am recording the fuller extent rather than raising the severity.
   The literals at `:77-80` are overwritten at import by
   `compute_rmc_spectrum_fractions()` (called at `:260`), and nothing then reads
   any of the four: `rmc_0n_normalization` and `rmc_1n_normalization` go through
   `rmc_spectrum_fraction()` instead, and `git grep RMC_SPECTRUM_FRAC` across
   `JobConfig/ensemble` finds no other consumer than `test_rmc_fractions.py`, which
   defines its own. So it is not only the 80 MeV pair — all four, plus the
   `compute_rmc_spectrum_fractions` function and its `global` statements, can go.

**Prior findings**

- 1. 🟠 [was S1] The `e_min ≤ 75` branch — **FIXED in `d61e3be`, verified.** Both
  functions now call `rmc_spectrum_fraction(e_min, RMC_KMAX_*, knockout=*)`, which
  is `plestid_integral(e_min)/plestid_integral(57)` — the relative form the
  prefactor requires. I checked the arithmetic rather than the diff: it returns
  exactly `1.000000` at `e_min = 57` for both modes, `0.0` above `k_max`, and there
  is no discontinuity left at 75 MeV. See the verified section for the agreement at
  the production default.
- 2. 🟠 [was S1] Duplicate constants — **UNADDRESSED**, carried as finding 2.
- 3. 🟡 [was S2] `plestid_integral` added but never called — **FIXED.** It is now
  the sole source of the spectrum fractions, and the two `k_max` values in
  `:68-69` (101.8667 and 95.4489 MeV) are the ones I recovered from the hardcoded
  constants last pass, to within the precision I quoted.
- 4. 🟡 [was S2] The four new tests in `test_normalizations.py` assert only `> 0`
  — **UNADDRESSED**; that file is untouched by the delta. Finding 4 above is a
  second instance of the same pattern.
- 5. 🟡 [was S2] `on_spill_time` scaled by an uncited `88/496` — **UNADDRESSED.**
- 6. 🟡 [was S2] `Stage2_build_sampler.sh`: the stray `--tape` on the RPCInternal
  list, and `check_file_lists` warning instead of exiting — **UNADDRESSED**;
  `Stage2_build_sampler.sh` is untouched by the delta.
- 7. ⚪ [was S3] Of the three small items: the dead `e_threshold = 57.0` in both
  new functions is **FIXED** (removed with the branch). The
  `make_template_fcl.py:63-73` aliases and the contradictory
  `Stage1_initiate_ensemble.sh:24-25` comment are **UNADDRESSED**, both files being
  untouched — though the `calculateEvents.py` side of the alias question is now
  resolved, which was @michaelmackenzie's point.

- 🟢 The new spectrum fraction reproduces the old correct branch at the production
  threshold, so this fix does not move any yield anyone has already computed. At
  `e_min = 80`, `plestid_integral(80)/plestid_integral(57)` gives `0.145039` for
  0N against the old `0.03319/0.22887 = 0.145017` — a relative difference of
  `1.5e-4` — and `0.021381` for 1N against `0.0013175/0.061620 = 0.021381`, to
  `1.7e-5`.
- 🟢 The two `k_max` values are consistent with all four literature fractions, not
  just the two the fix uses: computing `plestid_integral(E, k_max, k_max, n) /
  plestid_integral(0, k_max, k_max, n)` gives 0.228873 vs 0.22887, 0.061619 vs
  0.061620, 0.033195 vs 0.03319 and 0.0013175 vs 0.0013175 — the largest
  disagreement being 0.016%.
- 🟢 `rmc_spectrum_fraction` is well behaved outside the range that mattered
  before: `e_min` above `k_max` returns `0.0` rather than a negative or a divide
  error, and `e_min = 0` returns `4.369 = 1/0.22887`, which is the right answer for
  a quantity defined relative to E > 57.

**Validation check**
- Build/tests run: **none complete at this head.** `mu2e/buildtest` has been
  `pending` at `3c13c282` since 2026-08-20T15:52:51Z and FNALbuild has posted no
  table for it; the green table on this PR is from build 658 at `274b68d8`. I did
  not trigger another, since one is already queued — and it would not settle
  anything here, because the Production job list contains no python test target, so
  none of these files run in CI either way. My own checks were the arithmetic
  above and `git grep` for the dead constants.
- Config contract check: partial, unchanged from the last pass.
- Cross-repo consistency: n/a — no Offline or trig-config interface is touched.

**Residual risk**
- Whether removing `filter_efficiency` is physically correct is not something I can
  judge; I can only say it is a 6.2× change that is not described anywhere.
- The physics inputs themselves — the 0.099/0.901 knockout split, and the two
  `k_max` endpoints as values for Al-27 — remain outside what I can check. I
  verified internal consistency, not the literature.

**Author follow-ups**
1. Restore `filter_efficiency` in `rpc_normalization`, or state the reason for
   removing it and split it out (finding 1). This is the one item I would hold the
   merge for.
2. Reconcile `1.41e-5` against `RMC_GT_57_PER_CAPTURE = 1.43e-5` and drop the three
   duplicate constants (finding 2, carried).
3. Finish or revert the plain-RMC branch removal in `calculateEvents.py`
   (finding 3).
4. Make `test_rmc_fractions.py` import `plestid_integral` and assert, or rename it
   (finding 4).
5. Delete the four dead `RMC_SPECTRUM_FRAC_*` constants and
   `compute_rmc_spectrum_fractions` (finding 5).

sophiemiddleton:
Responses:

Author follow-ups

1    Restore filter_efficiency in rpc_normalization, or state the reason for
    removing it and split it out (finding 1). This is the one item I would hold the
    merge for.
    
    --- > this fixes a bug it needs to be removed
    
    Reconcile 1.41e-5 against RMC_GT_57_PER_CAPTURE = 1.43e-5 and drop the three
    duplicate constants (finding 2, carried).
    
    ---> can do only one is used 
    Finish or revert the plain-RMC branch removal in calculateEvents.py
    (finding 3).
    
    ---> no reason to do this, it might be useful to look over old data sets 
    Make test_rmc_fractions.py import plestid_integral and assert, or rename it
    (finding 4).
    
    --> we dont need this file I can remove 

    Delete the four dead RMC_SPECTRUM_FRAC_* constants and
    compute_rmc_spectrum_fractions (finding 5).

    ---> I can remove

oksuzian (changes requested):
### PR Review Summary — Mu2e/Production#567

Reviewed at head `fb6e2cac`. Re-review of `3c13c282`.

**Decision**
- 🔴 request changes

The delta since my last pass is a single commit, `fb6e2cac` ("test"), and it
touches only the four Stage shell scripts. None of the five findings from the
last review is addressed, so all five carry forward; @sophiemiddleton answered
each of them in a comment on 2026-08-20, and I have folded those answers into the
carry-forward list below rather than repeating the findings.

The new commit is not a fix commit, though. It adds a SAM version-discovery
stage, replaces the measured cosmic livetime with a hardcoded number, and makes
Stage 3 copy into a mu2epro-owned directory. Two of those are blockers on their
own, and none of it is mentioned in the PR description.

**Scope of this pass**
- `3c13c282..fb6e2cac`: new `Stage0_find_latest_sample.sh` (+96), and
  `Stage1_initiate_ensemble.sh` (+51/−20), `Stage2_build_sampler.sh` (+131/−14),
  `Stage3_configure_ensemble_campaign.sh` (+19/−1).
- The python files — `normalizations.py`, `calculateEvents.py`,
  `test_rmc_fractions.py`, `test_normalizations.py` — are untouched by the delta.

---

**New findings**

1. 🔴 [S0] **The cosmic livetime is now a hardcoded literal that silently overrides `--livetime`, and it no longer tracks the file list it is supposed to describe.**
   - Evidence: `Stage1_initiate_ensemble.sh:137-138` comments out the
     `printCosmicLivetime.fcl` call and sets
     `LIVETIME=5.72e6 # ... # 720575 FIXME this is hardcoded due to now reprocessing`.
     That assignment is unconditional and sits after the option loop, so the
     `--livetime` value parsed at `:42` is discarded without a word. `LIVETIME` then
     feeds every `calculateEvents.py --livetime` call at `:147-180` and is written
     to the config as `onspilltime` at `:195`, which Stage 2 reads back.
   - Impact: every background yield in the ensemble — DIO, IPA, RPC internal and
     external, all four RMC channels — is normalized to 5.72e6 s regardless of how
     many cosmic files were actually selected. With the new `--njobs` behaviour at
     `:122-128` the cosmic sample can be anything from a handful of files to the
     whole 5000-file dataset, and the backgrounds do not move with it. The number
     the code used to compute is recorded in the same comment as 720575, a factor
     of 7.9 away, so the two are not interchangeable.
   - Suggested fix: keep the measured path and make the override an explicit,
     documented option — restore the `printCosmicLivetime.fcl` call, and if a fixed
     value is needed while the MDC2025 bug is open, take it from `--livetime` and
     fail when neither is available. A committed literal with a FIXME is the thing
     that outlives the bug it was written for.

2. 🔴 [S0] **Stage 3 now copies into a directory only `mu2epro` can write, and exits 1 there — after the tarball has already been uploaded to tape and registered in SAM.**
   - Evidence: `Stage3_configure_ensemble_campaign.sh:338-344` (and the same block
     at `:138-144`) run
     `cp ${CAMPAIGN_JSON_FILE} /exp/mu2e/app/users/mu2epro/production_manager/poms_map/`
     and `exit 1` on failure. That directory is `drwxr-xr-x mu2epro mu2e`; I
     confirmed a write there is refused for an ordinary user. The second copy sits
     after the `mu2eFileUpload` and the `samweb add-file-location` at `:240`.
   - Impact: any user who is not mu2epro gets a hard failure at the end of a run
     that has already pushed a cnf tarball to tape and registered its location in
     SAM. Re-running after fixing the permission means re-uploading under the same
     cnf name, which is exactly the situation the "never reuse a cnf name" rule
     exists for.
   - Suggested fix: make the destination an option with a default, and warn rather
     than exit when the copy is not possible — the script already prints the
     `--jobdefs` path the user needs, so the copy is a convenience, not a
     precondition.

3. 🟠 [S1] **`RMCVERSIONINT` and `RMCVERSIONEXT` are shared between the 0N and 1N discovery blocks, so the 1N versions are never looked up when 0N is enabled.**
   - Evidence: `Stage2_build_sampler.sh:189` discovers `RMCVERSIONINT` from
     `RMCPhaseSpace0NInternal`; `:211` guards the 1N Internal lookup with
     `[[ -z "${RMCVERSIONINT}" ]]`, which is false by then. The same holds for
     `RMCVERSIONEXT` at `:200` and `:222`. Both variables are then used for the 1N
     datasets at `:303-304`, `:401-402` and `:463-464`. Stage 1 sets both
     `INCLUDE_RMCN0` and `INCLUDE_RMCN1` to 1 by default, so this is the default
     path.
   - Impact: the 1N datasets are built with whatever version 0N resolved to. Today
     that is harmless — I checked SAM and 0N and 1N happen to share their latest
     versions (`au` internal, `at` external for both) — but the whole purpose of
     the discovery step is to cope with them diverging, and when they do the result
     is a valid-looking dataset name for the wrong version. The validation step at
     `:371` will not catch it, because the wrong name still resolves to a real
     dataset with files.
   - Suggested fix: give 0N and 1N their own four variables, and drop the
     `[[ -z ... ]]` guards that were meant to preserve a manual `--rmcversionint`
     override but now also suppress the second lookup.

4. 🟡 [S2] **Two Stage 1 options are now inert.** `--version` is still parsed at
   `Stage1_initiate_ensemble.sh:63` but `VERSION` is read nowhere else in the file
   since `primaries=` was dropped from the config output; and `TAG` defaults to
   empty with no check, so a run without `--tag` writes its configuration to a file
   literally named `.txt` and only fails later, in Stage 2. Delete the `--version`
   case, and validate `TAG` the way `Stage0_find_latest_sample.sh:60` validates
   `--defname`.

5. ⚪ [S3] **`--tape` has been added to the `IPAMuminusMichel` file list**
   (`Stage2_build_sampler.sh:406`), alongside the one on `RPCInternal` that I
   raised last time and that is still there. If both are deliberate, a comment
   saying why those two datasets are read from tape would stop the next reader
   removing them. The commit message "test" is also worth rewriting before merge.

**Carried forward from `3c13c282` — all five still open**

- 1. 🟠 [was S0] `filter_efficiency` deleted from `rpc_normalization`. **PARTIAL.**
  @sophiemiddleton has answered that this is deliberate ("this fixes a bug it
  needs to be removed"), which resolves the question I could not answer myself, so
  I am lowering this from S0 to S1 and no longer treating it as possibly
  accidental. What remains open is the record: `normalizations.py:535` still
  computes `filter_efficiency` and nothing reads it, and a 6.2× change to the RPC
  background is still described nowhere in the PR body or the commit messages.
  Delete the dead line and add a sentence to the description saying what the bug
  was.
- 2. 🟠 [was S1] `RMC_BR_MUON_CAPTURE`, `RMC_RATE_GT_57 = 1.41e-5` and
  `RMC_INTERNAL_EXTERNAL_RATIO` still duplicate `constants.py`, and 1.41e-5 still
  disagrees with the sourced `RMC_GT_57_PER_CAPTURE = 1.43e-5`
  (`normalizations.py:62-63,85`). **UNADDRESSED.** The answer was "can do only one
  is used" — but both are used, by different functions, which is the finding.
- 3. 🟡 [was S2] The plain-RMC branch in `calculateEvents.py` computes a yield and
  prints nothing. **UNADDRESSED**, and the answer was that the branch is worth
  keeping for old datasets. If it is kept, it needs its `print` back, or it is kept
  in a form that cannot be used.
- 4. 🟡 [was S2] `test_rmc_fractions.py` duplicates `plestid_integral` and asserts
  nothing while being collected as a test. **UNADDRESSED**; the answer was "we
  dont need this file I can remove".
- 5. 🟡 [was S2] The four `RMC_SPECTRUM_FRAC_*` constants and
  `compute_rmc_spectrum_fractions` are dead (`normalizations.py:77-80,222-246`).
  **UNADDRESSED**; the answer was "I can remove".
- Older items still open and unchanged: the `> 0`-only assertions in
  `test_normalizations.py`; the uncited `88/496` on `on_spill_time`;
  `check_file_lists` warning instead of exiting (`Stage2_build_sampler.sh:418,480`
  — this file is touched by the delta now, so it is in scope again); and the
  `make_template_fcl.py:63-73` aliases.

- 🟢 `Stage0_find_latest_sample.sh` works as documented. Run against SAM it returns
  `ap` for `dts.mu2e.CosmicSignal%.MDC2025%.art`, and
  `samweb count-definition-files dts.mu2e.CosmicSignal.MDC2025ap.art` is 5000, so
  the version it picks is a populated dataset and not an empty definition. The
  same holds for the RPC, IPA and RMC patterns the Stage 2 discovery uses.
- 🟢 A failed discovery does not reach submission. The `DISCOVERY_FAILED` path only
  warns (`:239-251`), but an empty version produces a malformed dataset name that
  the availability check at `:287-365` catches, and that block does `exit 1` at
  `:371`. This is what keeps finding 3 at S1 rather than S0 — but note it does not
  help there, where the wrong name is a real dataset.
- 🟢 `MDC2025aw`, the new `CURRENT` in Stage 2 and `VERSION` in Stage 3, exists as
  a SimJob Musing on cvmfs, so the `setup.sh` path in `Stage2_build_sampler.sh:26`
  resolves.
- 🟢 The config-file contract between Stage 1 and Stage 2 survives the rename of
  the RMC keys: every key Stage 2 sources at `:127-148` (`DEM_emin`, `RPC_TMIN`,
  `RPC_emin`, `RMC_N0_emin`, `RMC_N1_emin`, `RMC_kmax`, `IPA_emin`) is written by
  Stage 1 at `:197-207`, and `primaries`, which Stage 1 stopped writing, is read
  nowhere.

**Validation check**
- Build/tests run: `mu2e/buildtest` is pending at `fb6e2cac`, triggered at
  2026-09-09T16:09Z. I did not trigger another. It would not settle any of this in
  any case: the Production job list has no python or shell test target, so none of
  these files runs in CI. My checks were the SAM queries and permission test quoted
  above, plus reading the scripts.
- Config contract check: pass for the Stage 1 → Stage 2 config file (see the last
  🟢); fail for `--livetime`, which is parsed and then ignored.
- Cross-repo consistency: n/a — no Offline or trig-config interface is touched.

**Residual risk**
- Whether 5.72e6 s is the right livetime for any particular cosmic dataset is not
  something I can check; I can only say it is fixed, that it overrides the option
  meant to set it, and that it disagrees with the number in its own comment.
- The physics inputs — the 0.099/0.901 knockout split and the two `k_max`
  endpoints — remain outside what I verified, as last time.

**Author follow-ups**
1. Restore the measured livetime, or route the fixed value through `--livetime` so
   the option is not silently discarded (finding 1).
2. Make the Stage 3 copy optional and non-fatal, and take the destination from an
   option (finding 2).
3. Split `RMCVERSIONINT`/`RMCVERSIONEXT` into separate 0N and 1N variables
   (finding 3).
4. Apply the four cleanups you already agreed to in your 2026-08-20 comment —
   the duplicate constants, the plain-RMC `print`, `test_rmc_fractions.py`, and
   the dead `RMC_SPECTRUM_FRAC_*` block — and delete the now-unused
   `filter_efficiency` line.
5. Please update the PR description: it still describes only the RMC normalization
   work, and the branch now also carries a SAM version-discovery stage, a livetime
   change and a production-directory copy. Splitting the shell-script work into its
   own PR would make both halves reviewable; at minimum the description should say
   what is in here.

oksuzian (approved):

PR #568, opened 2026-08-15
Calo DTS cluster filtering in mixing

michaelmackenzie:
Add the calo DTS cluster filter module to the mixing workflow, off by default. In Run 1B we can turn this on for the No Primary mixing to only digitize and save events that have high energy deposits. This will allow us to simulate higher statistics without significant CPU or data output increases.

### DIFF
--- a/JobConfig/mixing/NoPrimary.fcl
+++ b/JobConfig/mixing/NoPrimary.fcl
@@ -1,11 +1,11 @@
 #
-# configuration for NoPrimary: in this case, Untriggered selection should be wide open (null prescaler)
+# configuration for NoPrimary: in this case, Untriggered selection should be wide open (null prescaler) except for digitization success
 #
 physics.filters.Triggerable : {
   module_type : Prescaler
   prescaleFactor : 1
   prescaleOffset : 0
 }
-# turn off any filtering for triggerable output stream
+# turn off any filtering for triggerable output stream, other than successful digitization
 outputs.Output.RejectEvents : [ ]
 outputs.Output.SelectEvents : [ DigitizePath ] # Don't save events that fail digitization
--- a/JobConfig/mixing/NoPrimary.fcl
+++ b/JobConfig/mixing/NoPrimary.fcl
@@ -7,5 +7,5 @@ physics.filters.Triggerable : {
   prescaleOffset : 0
 }
 # turn off any filtering for triggerable output stream
-outputs.Output.RejectEvents : []
-outputs.Output.SelectEvents : []
+outputs.Output.RejectEvents : [ ]
+outputs.Output.SelectEvents : [ DigitizePath ] # Don't save events that fail digitization
--- a/JobConfig/mixing/prolog.fcl
+++ b/JobConfig/mixing/prolog.fcl
@@ -81,13 +81,26 @@ Mixing : {
       ReturnValue : true
     }
 
+    CaloDtsClusterFilter : {
+      module_type : CaloDtsClusterFilter
+      CaloShowerSteps        : ["compressDetStepMCs", "MuStopPileupMixer", "EleBeamFlashMixer", "MuBeamFlashMixer", "NeutralsFlashMixer"]
+      MinimumStepEnergy      : 0.
+      MinimumStepTime        : 400.
+      MaximumStepTime        : 1.e6 # can't do fmod(time, 1695) accurately at high times
+      MinimumClusterEnergy   : 50.
+      TimeWindow             : 100.
+      SpaceWindow            : 80.
+      NullFilter             : true # Default to off
+      DiagLevel              : 0
+    }
+
   }
   StepMixSequence : [  MuStopPileupMixer, EleBeamFlashMixer, MuBeamFlashMixer, NeutralsFlashMixer ]
   StepFilterSequence : [ MuStopPileupFilter, EleBeamFlashFilter, MuBeamFlashFilter, NeutralsFlashFilter ]
   # when mixing early events the intensity must be scaled down
   EarlyReductionFactor : 1.0e-2
 }
-Mixing.PileupMixSequence : [ PBISim, DTSFilter, @sequence::Mixing.StepMixSequence, @sequence::Mixing.StepFilterSequence ]
+Mixing.PileupMixSequence : [ PBISim, DTSFilter, @sequence::Mixing.StepMixSequence, @sequence::Mixing.StepFilterSequence, CaloDtsClusterFilter ]
 Mixing.MixSequence : [ @sequence::Mixing.PileupMixSequence, @sequence::Digitize.DigitizeSequence ]
 # override the Trigger 'PrepareDigis' sequence; the mixers must preceed the digi making
 TrigRecoSequences.cfoDataGen: [ @sequence::Mixing.PileupMixSequence, @sequence::Digitize.DigitizeSequence, ProcessCFOData ]
--- a/JobConfig/mixing/prolog.fcl
+++ b/JobConfig/mixing/prolog.fcl
@@ -84,12 +84,12 @@ Mixing : {
     CaloDtsClusterFilter : {
       module_type : CaloDtsClusterFilter
       CaloShowerSteps        : ["compressDetStepMCs", "MuStopPileupMixer", "EleBeamFlashMixer", "MuBeamFlashMixer", "NeutralsFlashMixer"]
-      MinimumStepEnergy      : 0.
-      MinimumStepTime        : 400.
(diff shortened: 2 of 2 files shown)


oksuzian (approved):
### PR Review Summary — Mu2e/Production#568

Reviewed at head `e15d665ccda22b45dddcf8684d0bd956e9371460`. First pass.

**Decision**
- 🟢 approve

The "off by default" claim holds — I traced every module on `DigitizePath` to
confirm nothing in it can reject once `NullFilter : true` is set, so this merges
inert. Findings are a stale comment and one question about reach.

**Scope understood**
- Adds `CaloDtsClusterFilter` to `Mixing.PileupMixSequence` with `NullFilter : true`,
  so it accepts every event until someone turns it off.
- Switches `NoPrimary.fcl`'s `outputs.Output.SelectEvents` from `[]` to
  `[ DigitizePath ]`, which is the plumbing that makes the filter's rejection
  actually drop the event from the output once it is enabled.

**Findings**

1. 🟡 [S2] The comment above the changed lines now says the opposite of what they do
   - `NoPrimary.fcl:9` still reads `# turn off any filtering for triggerable output
     stream`, directly above `SelectEvents : [ DigitizePath ]`, which is filtering.
     The two comments on those lines now contradict each other. Replacing line 9
     with something like "only save events that complete digitization; see
     Mixing.CaloDtsClusterFilter" would carry the intent. (The file header's
     "Untriggered selection should be wide open (null prescaler)" is still accurate
     — that describes the `Triggerable` prescaler, which this PR does not touch.)

2. 🟡 [S2] The filter also lands in the trigger CFO sequence — intended?
   - `prolog.fcl:105` rebuilds `TrigRecoSequences.cfoDataGen` from
     `@sequence::Mixing.PileupMixSequence`, so appending `CaloDtsClusterFilter` to
     that sequence puts it in the CFO data-generation path as well as
     `DigitizePath`. Inert today, but when it is enabled for Run 1B No Primary it
     will gate that path too, and anyone doing trigger studies on those files gets
     the calo-filtered subset without an obvious signpost. If that is wanted, worth
     a line in the prolog saying so; if not, the filter belongs in `MixSequence`
     rather than `PileupMixSequence`.

3. ⚪ [S3] The tuning values carry no units or provenance
   - `MinimumStepTime : 400.`, `MinimumClusterEnergy : 50.`, `TimeWindow : 100.`,
     `SpaceWindow : 80.` — the module's own `Comment` strings do not give units
     either, so nothing in either repo says whether 50 is MeV or where it came
     from. These will decide what gets kept in a large production sample; one
     trailing comment per line, in the style of the `MaximumStepTime` line right
     above them, is enough. (`MinimumStepEnergy : 0.` also just restates the
     module's own default — harmless, but it can go.)

- 🟢 The FHiCL contract matches the module exactly. All nine keys correspond to
  entries in `CaloDtsClusterFilter::Config`
  (`Offline/Filters/src/CaloDtsClusterFilter_module.cc:44-54`), which uses the
  validated `art::EDFilter::Table<Config>` pattern. `CaloShowerSteps` is a
  `Sequence<art::InputTag>`, `NullFilter` a `bool` defaulting to `false`, and
  "if true, do not apply any filtering and accept all events" is what the
  `# Default to off` comment claims.
- 🟢 All five input tags resolve at the point the filter runs. The four mixer
  labels are exactly `Mixing.StepMixSequence`, which precedes the filter in
  `PileupMixSequence`; `compressDetStepMCs` comes from the input file, as the
  `caloShowerStepMixer` mixing map on `prolog.fcl:32` already reads that same
  label for `CaloShowerStep`s. This matters because `fillStepsList` uses
  `getValidHandle`, which throws rather than skipping a missing tag — a wrong
  label here would be a hard failure the moment the filter is switched on.
- 🟢 The enable switch is genuinely free. `filter()` tests `nullFilter_` and
  returns before touching the event
  (`CaloDtsClusterFilter_module.cc:304-309`), so no clustering work is done in
  the default configuration.
- 🟢 `SelectEvents : [ DigitizePath ]` changes nothing today. `DigitizePath` is
  `Mixing.MixSequence` = `PileupMixSequence` + `Digitize.DigitizeSequence`.
  Walking it: `PBISim` and everything in `DigitizeSequence`
  (`CommonMC/TrackerMC/CaloMC/CrvDAQ` `DigiSim`, `compressDigiMCs`) are
  producers; `StepFilterSequence` members are `MixingFilter` **producers**, not
  filters; `DTSFilter` is a `FixedFilter` with `ReturnValue : true`; and
  `CaloDtsClusterFilter` is null-filtered. The `Triggerable*` `StrawDigiMCFilter`s
  that can reject live on the `Triggerable*Path`s, not on `DigitizePath`. Only the
  four `MixBackgroundFrames` mixers can return false, and only if they run out of
  pileup to supply.

**Validation check**
- Build/tests run: CI is **pending** at this head — `mu2e/buildtest` was triggered
  when the PR opened and has not reported. A run is already in progress so I did
  not trigger another. Note that `mu2e/buildtest` on `Production` does not exercise
  a mixing job, so a green will confirm the fcl parses in the jobs it does run,
  not that the new filter behaves.
- Config contract check: pass — key-by-key against `Config`, and every input tag
  traced to its producer. See the verified section.
- Cross-repo consistency: pass. `CaloDtsClusterFilter_module.cc` is already on
  `Offline` `main` and registered in `Filters/CMakeLists.txt`, so no companion PR
  is outstanding.

**Residual risk**
- I checked that the filter is wired correctly and is inert as configured. I did
  not run a mixing job with `NullFilter : false`, so the actual acceptance — and
  whether 50/100/80 give the intended rate — is unverified here.

**Author follow-ups**
1. Update `NoPrimary.fcl:9` (finding 1).
2. Confirm whether gating `TrigRecoSequences.cfoDataGen` is intended (finding 2).
3. Add units to the four cut values (finding 3).

michaelmackenzie:
Responding to the AI comments:
1. I updated the comments
2. Yes, it's intentional to not process anything (including the CFO data) if we're filtering out the event
3. I added comments for units

oksuzian (approved):
### PR Review Summary — Mu2e/Production#568

Reviewed at head `7941e15749e6ef95771dfaf97a70a102f192615c`. Re-review of `e15d665c`; the delta is comment-only (`+8-8` across the two files).

**Decision**
- 🟢 approve

All three items from the previous pass are addressed. Nothing new.

**Prior findings**

1. 🟢 [was S2] Stale comment on `NoPrimary.fcl` — **FIXED in `7941e157`, verified.**
   Line 9 now reads "turn off any filtering for triggerable output stream, other
   than successful digitization", and the file header gained "except for
   digitization success". One thing to keep in mind rather than change: once
   `NullFilter` goes to `false`, `DigitizePath` also fails on a calo-cluster
   rejection, which is not a digitization failure — so the comment describes the
   inert configuration accurately and the enabled one loosely. Not worth another
   edit now; worth a word when the switch is flipped.

2. 🟢 [was S2] Filter reaching `TrigRecoSequences.cfoDataGen` — **ANSWERED,
   withdrawn.** *"Yes, it's intentional to not process anything (including the CFO
   data) if we're filtering out the event."* That is a design decision, and a
   sensible one — the CPU saving is the point of the PR, and processing CFO data
   for an event that will not be written would undercut it. The rationale now
   lives only in this thread, not in `prolog.fcl`, which is fine by me.

3. 🟢 [was S3] Units on the cut values — **FIXED in `7941e157`, verified.** All six
   lines carry units, and each matches what the module actually does: energies in
   MeV against `CaloShowerStep::energyDepBirks()`, times in ns against
   `css.time()` wrapped on `getNominalDRPeriod()`, and `SpaceWindow` in mm against
   the (x,y) step positions. Where the values themselves came from is still
   unrecorded — not asking for that, they are tuning parameters you own.

**Findings**
- None new. The delta touches only comments; the module wiring, input tags and
  `SelectEvents` semantics I verified at `e15d665c` are unchanged, so those checks
  carry over.

**Validation check**
- Build/tests run: `mu2e/buildtest` **passed at `e15d665c`** and is **pending at
  `7941e157`**. A run is already in progress, so I did not trigger another. Since
  the delta is comment-only FHiCL, the previous green covers the substance.
- Config contract check: pass, carried over from the previous pass — no key,
  value or sequence changed, only trailing comments.
- Cross-repo consistency: pass, unchanged.

**Residual risk**
- Same as before: the filter is verified as correctly wired and inert as
  configured. Nobody has yet run a mixing job with `NullFilter : false`, so the
  acceptance those six numbers produce is still unmeasured.

**Author follow-ups**
- None.

oksuzian:
@FNALbuild run build test

